PR #16, opened 2014-11-14
[JENKINS-25588] - Migrate vSphereCloud to Credentials plugin

oleg-nenashev:
<b>Warning!</b> This is a breaking change. A new major version (or a significant modification) are required

Signed-off-by: Oleg Nenashev o.v.nenashev@gmail.com


chris5287:
I've just updated to 2.0 plugin but I cannot figure out how to setup the credential store correctly. No matter what I try I get "Username is not specified" when testing the connection, and any VM slave I try to start returns the following errors:
[JENKINS-...] Starting Virtual Machine...
[JENKINS-...] EXCEPTION while starting VM
[JENKINS-...] java.net.MalformedURLException: no protocol: null/sdk

jswager:
My initial test of this pull request SEEMED to have worked, but when re-ran the tests - i'm getting the same error.

@oleg-nenashev Can you confirm the behavior described here?  And if so, do you know how to resolve it?

oleg-nenashev:
Fixed the validation handler.
I'm working on the second issue. As I remember, the connection worked for me, but it definitely makes sense to re-check it

PR #18, opened 2015-01-05
[JENKINS-25588] - Fixed the "Test Connection" validation handler

oleg-nenashev:
Fixes the error with vSphere validation handlers (caused by #16 )

Signed-off-by: Oleg Nenashev <o.v.nenashev@gmail.com>

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/vSphereCloud.java
+++ b/src/main/java/org/jenkinsci/plugins/vSphereCloud.java
@@ -153,7 +153,7 @@ public class vSphereCloud extends Cloud {
 	}
     
     public VSphere vSphereInstance() throws VSphereException{
-    	return VSphere.connect(vsHost + "/sdk", username, getPassword());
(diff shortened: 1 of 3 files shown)


jswager:
@oleg-nenashev This resolves the "Test Connection" button from the main Jenkins config screen.  But this doesn't solve the "no protocol: null/sdk" error when a slave is being started.  

Quick look at the debugger: in vSphereCloud.java, in the vSphereInstance() method - both vsHost and username are null - which results in an invalid URL, which might explain this problem.

oleg-nenashev:
@jswager 
Yes, my bad. I have not forced null values in <code>readResolve()</code> and I missed such issues during the testing of the initial PR. In new commits I removed all usages of all variables. Hopefully, the change works now.

jswager:
@oleg-nenashev This last round of changes does seem to fix everything.  The test buttons are working from the main config, slave config, and build steps.  A slave can be powered on now.

Thanks for you quick responses and effort on getting these issues resolved so quickly.

PR #23, opened 2015-01-29
ClusterSelection: Get datastores directly from cluster

vorbis:
In our Vsphere setup, searching for Datastores under the cluster doesn't work.
When calling getDataStoreClusters on a cluster, the cluster is found though.
Unfortunately I don't know enough about the vmware vim25 api to know why the
datastore wasn't found.
It is found if I don't pass in the cluster for the searching and the searching
takes  place on the root entity.
Pass in logger to Vsphere clone method so that output can be seen in jenkins
console.

While testing this I realized that several arguments are not expanded.
Cluster for instance is not expanded in either Clone or Deploy.
This would mean that ${CLUSTER} for instance wouldn't be expanded.
Improved logging messaging as well so errors in configuration are easier to find.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/vsphere/builders/Clone.java
+++ b/src/main/java/org/jenkinsci/plugins/vsphere/builders/Clone.java
@@ -84,7 +84,6 @@ public class Clone extends VSphereBuildStep {
 
 	private boolean cloneFromSource(final AbstractBuild<?, ?> build, final Launcher launcher, final BuildListener listener) throws VSphereException {
 		PrintStream jLogger = listener.getLogger();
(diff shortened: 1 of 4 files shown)


grjones:
@damienbiggs I believe this change broke some things.  My vSphere doesn't have any clusters configured.  I was using the Datacenter name before, but after upgrading from 2.2 to 2.4 this no longer works.

damienbiggs:
Sorry about that.
The old approach wasn't working for me in our Vcenter with clusters, I'll try and create a patch on Monday to incorporate both approaches. Would you be able to test a merge request, ideally it'd be good to know it works before merging it to master.

grjones:
@damienbiggs Absolutely I can test. Thanks!

PR #24, opened 2015-01-30
VsphereSelection: Allow dynamic selection of vsphere cloud

vorbis:
A jenkins user could have a job that can be used for two separate VSphere
instances.
At present there is no way to select the vsphere instance using a build parameter.
This adds a parameter called ${VSPHERE_CLOUD_NAME} as an vsphere cloud option for selection.
If selected then the jenkins user can add a parameter of the same name for the
build.


vorbis:
Btw, is there a formatting jar for intellij for the project?
My formatting at times seems at odds with the existing files.
Thanks.

jswager:
No formatting jar that I'm aware of.  From what I've seen, the contributors to the plugin use a variety of editors.

PR #28, opened 2015-04-26
ClusterFix: Don't assume that the cluster should exist

damienbiggs:
For cloning / deplying a vm, a cluster might not exist.
Use old datastore check first before checking the cluster resource.
I made a wrong assumption that the cluster should exist.


damienbiggs:
@grjones

Can you checkout this branch, build it and verify if it fixes your issue?
I think it should as it won't throw an exception for the cluster not being found any more.

damienbiggs:
@grjones
Were you able to verify this pull request?
Regardless I think this request should be merged as it should fix scenarios where people were using the datacenter as the cluster.

grjones:
Hi @damienbiggs, unfortunately I don't have access to that vsphere any longer and the ones I do have access to now are clustered.  Code-wise it looks like it should work if that helps at all. :/

damienbiggs:
Thanks @grjones 
Yeah I agree, this request should fix the behavior you experienced.

PR #30, opened 2015-05-18
Logging: Log stack trace if exception message is null

damienbiggs:
I've noticed that sometimes the power on task fails with a null exception
message.
Log stack trace if the exception message is null.
Moved vsphere disconnect to a finally so that disconnect happens after an
exception  as well.


damienbiggs:
@jswager @ParagDoke 
Would it be possible to get commit permissions for the repo so that I could push some of the pull requests if they pass the build check.

Thanks.

jswager:
Working on the commits now...

damienbiggs:
Thanks @jswager

PR #33, opened 2015-06-14
added possibility of creating clone without specifying resource pool.

Ichai5:
Hi vsphere-cloud-plugin team,
 
I’m currently need that the cloned VM will be created on the same host as the source.
When I browse the vmware documentation on VirtualMachineRelocateSpec I see that if resource pool is not specified and host also, the current host is used.
So If I have understood, enable cloning the VM without specifying the Resource Pool will clone it to the same host.
For a clone operation from a template to a virtual machine, this pool is required according to the  vmWare Doc. 
So we can enable cloning a VM without specifying a resource pool when the source is not a template.
Please find the changeset that let me get this feature.
Are these changes acceptable for you ?
If yes can you add them to the next plugin version.
 
Thanks in advance,
 
Ichai


.project
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
daniel-beck: This file should not be in this commit.

src/main/java/org/jenkinsci/plugins/vsphere/builders/Clone.java
```diff
@@ -131,13 +134,6 @@ public FormValidation doCheckClone(@QueryParameter String value)
 			return FormValidation.ok();
 		}
 
-		public FormValidation doCheckResourcePool(@QueryParameter String value)
```
daniel-beck: Can't this be improved to check only if applicable?

src/main/java/org/jenkinsci/plugins/vsphere/tools/VSphere.java
```diff
@@ -212,7 +211,8 @@ private VirtualMachineRelocateSpec createRelocateSpec(PrintStream jLogger, boole
             }
             rel.setDatastore(datastore.getMOR());
         }
-        return rel;
+
```
daniel-beck: Unnecessary whitespace changes.

daniel-beck:
Please note that I'm not a developer of this plugin, so my review comments are just offered as advice, nothing more.

.project
Ichai5: You are right,
How can I remove from the pull request ?

jswager:
@daniel-beck Maybe not a maintainer yet - want to be?  Pretty easy to fix that.  Your ongoing input has been valuable.

.project
jswager: Make a commit to Ichai5:master that deletes the file.  The PR will be automatically updated when the change is pushed to GitHub Ichai5:master.

daniel-beck:
@jswager Thanks, but my plate is full, and I don't use vSphere anyway, so I doubt I'd make a good maintainer. I'm still available if you have specific questions though.

jswager:
@daniel-beck Reviewing pull requests is always a great way to help!  Thanks!

PR #35, opened 2015-09-09
Dynamic On-Demand vSphere Slave

ghost1874:
This is an initial design for configuring jenkins to dynamically create vsphere clones to use as nodes.   During "Configure Jenkins" you can now add slave templates that define how new clones can be provisioned when a job needs a node and no node is able to accept the job.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/vSphereCloudSlave.java
+++ b/src/main/java/org/jenkinsci/plugins/vSphereCloudSlave.java
@@ -327,7 +327,7 @@ public class vSphereCloudSlave extends Slave {
 
                 // If this is the last run in a limited test run, then flag the node as offline.
                 if (NumberOfLimitedTestRuns >= LimitedTestRunCount) {
-                    doingLastInLimitedTestRun = Boolean.TRUE;
+                    doingLastInLimitedTestRun = Boolean.FALSE;
                 }
             } else {
                 vSphereCloud.Log(listener, "Terminating build due to limited build count: %d", LimitedTestRunCount);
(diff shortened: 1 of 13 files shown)


jswager:
Pretty impressive work.  @lfeagan - are you concerns addressed such that this code could be merged?

src/main/java/org/jenkinsci/plugins/vSphereCloudSlave.java
```diff
@@ -113,9 +152,9 @@ public boolean isLaunchSupportForced() {
         return ((vSphereCloudLauncher) getLauncher()).getOverrideLaunchSupported() == Boolean.TRUE;
     }
 
-    public void setLaunchSupportForced(boolean slaveLaunchesOnBootup) {
-        ((vSphereCloudLauncher) getLauncher()).setOverrideLaunchSupported(slaveLaunchesOnBootup ? Boolean.TRUE : null);
-    }
+//    public void setLaunchSupportForced(boolean slaveLaunchesOnBootup) {
```
lfeagan: Perhaps remove this unused comment block. There are a few others.

lfeagan:
@jswager My concerns have been addressed.

jswager:
@ghost1874 - This PR will need to be udpated to handle the merges from previous PRs.

ghost1874:
I looked at the build output.  I think it is still complaining on VSphere. java.   I have pulled that from jenkinsci/vsphere-cloud-plugin, accepted my changes, committed the changes, and pushed those changes back to ghost1874/vsphere-cloud-plugin.  If there is something else I need to do in-order to get this build fixed, please let me know.

PR #54, opened 2016-09-16
Implement JENKINS-38269

pjdarton:
Enhances existing vSphere cloud-template slave functionality to allow the launcher method to be configurable (allowing one to have a Java Web Start slave instead of SSHLauncher, for example), and/or to choose a a retention-policy of "use until it's idle" as an alternative to "use once".


segator:
could you add an idle timeout? as a retention-policy?

Example:
wait to destroy slave 20 min if is in idle, this allow an other job to reuse slave and avoid to deploy an other machine. 
if the slave is idle 20 min automatically be destroyed.

pjdarton:
Yes you can - That's what the (new) VSphereCloudRetentionStrategy does :-)

Prior to this enhancement, the only option available for cloud provisioned
slaves was the
RunOnceCloudRetentionStrategy which only runs a single build on the slave
before destroying it.
With this enhancement, you now have the option of choosing a
VSphereCloudRetentionStrategy
which keeps the slave alive after the build and only destroys it if it
remains idle for more than a period of time.
I chose a default of 5 minutes idle time, but you could configure it to be
20 minutes if you prefer.

PR #61, opened 2016-11-16
VSphere folder parameter

maleix:
A new folder parameter has been added for templates and clone/deploy
actions.

This parameter creates the new VM in the specified folder of vSphere
(which can be a path of folders or a unique name folder).

If left blank or folder not available, it will create the VM in the same
folder as the original VM.

It is specially interesting for organizing and separating templates from
autodeployed slaves.
For example, we have daily backups and we want separate folders for
templates and autodeployed dynamic slaves, so we can specify to backup
only the folders that contain the templates and omit folders that
contain the dynamic slaves.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
+++ b/src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
@@ -89,6 +89,7 @@ public class vSphereCloudSlaveTemplate implements Describable<vSphereCloudSlaveT
     private final String cluster;
     private final String resourcePool;
     private final String datastore;
+    private final String folder;
     private final String customizationSpec;
     private final String templateDescription;
     private int templateInstanceCap;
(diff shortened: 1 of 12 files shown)


segator:
I just give a try to your PR and works perfect!

pjdarton:
@maleix @jswager src/main/resources/org/jenkinsci/plugins/vsphere/builders/Clone/config.jelly looks incorrect.
The <f:entry...> with attribute title="${%Folder}" has field="datastore", but the Clone.java's folder field is actually called "folder".
(looks like this was a copy/paste from the <f:entry...> title="${%Datastore}" immediately above it)

Also, src/main/resources/org/jenkinsci/plugins/vsphere/builders/Deploy/config.jelly specifies field="Folder" instead of "folder" (that'll probably work, but it's not pretty).

(Apologies for the delay in noticing this; it's been 6 months since I last followed this plugin but I'm catching up now)

pjdarton:
I've raised https://github.com/jenkinsci/vsphere-cloud-plugin/pull/72/ as a prospective fix for these issues.

PR #63, opened 2016-12-22
Add new parameter 'ignoreIfNotExists'

igreenfield:
When try to powerOff non exists VM allow to ignore it.


igreenfield:
@jswager Could you check this pull request? I have 2 more changes that need this pull before.
thanks

PR #64, opened 2017-01-03
Add new parameter 'waitForIp4' And Prefix could be ${} (parameter)

igreenfield:


igreenfield:
@jswager Could you check this pull request?

PR #65, opened 2017-01-18
Fixed JENKINS-36952 and added timeOutInSeconds to Clone and Deploy

MarcelMue:
All changes were done for pipeline support only.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/vsphere/builders/Clone.java
+++ b/src/main/java/org/jenkinsci/plugins/vsphere/builders/Clone.java
@@ -39,6 +39,8 @@ import com.vmware.vim25.mo.VirtualMachineSnapshot;
 
 public class Clone extends VSphereBuildStep {
 
+    private final int TIMEOUT_DEFAULT = 60;
+
     private final String sourceName;
     private final String clone;
     private final boolean linkedClone;
@@ -48,13 +50,13 @@ public class Clone extends VSphereBuildStep {
     private final String folder;
     private final String customizationSpec;
     private final boolean powerOn;
-    private final int timeoutInSeconds;
+    private Integer timeoutInSeconds = new Integer(TIMEOUT_DEFAULT);
     private String IP;
 
     @DataBoundConstructor
     public Clone(String sourceName, String clone, boolean linkedClone,
                  String resourcePool, String cluster, String datastore, String folder,
-                 boolean powerOn, int timeoutInSeconds, String customizationSpec) throws VSphereException {
+                 boolean powerOn, Integer timeoutInSeconds, String customizationSpec) throws VSphereException {
         this.sourceName = sourceName;
         this.clone = clone;
         this.linkedClone = linkedClone;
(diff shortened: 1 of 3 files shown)


MarcelMue:
Is there a chance that this can be merged? @jswager

src/main/java/org/jenkinsci/plugins/vsphere/builders/Clone.java
```diff
@@ -48,21 +50,26 @@
     private final String folder;
     private final String customizationSpec;
     private final boolean powerOn;
+    private Integer timeoutInSeconds = new Integer(TIMEOUT_DEFAULT);
```
elordahl: • Why don't you declare this as `final` here and set it in the constructor if `null`?
• Why are you using an `Integer` over a primitive `int`?

phucyall:
Can this be fixed and merged please? The plugin is extremely broken in the pipeline with hard-coded timeout and the exceptions it causes. This PR would fix all of it, but it's been sitting here stale for 6 months.
It would close https://issues.jenkins-ci.org/browse/JENKINS-36952 if it was merged!

pjdarton:
@phucyall FYI at present this plugin is in a state of statis because the sole maintainer who has write access (which is required to merge a PR) is no longer able to spend time on this.  There are efforts being made (see #76 for discussion) to transfer access so we can get things moving again.

Note: at present *this* PR isn't in a good state (the issues raised above have not been dealt with) so it isn't presently in a state where I'd merge it even if I did have write-access and I expect that Eric would be like minded.

However, you could help...
If you were to create your own PR (and mention this one in the description/discussion) which implemented this fix in an acceptable way (i.e. addressing the issues that Eric raised above) then, personally, I'd be perfectly happy to merge it along with everything else (if I ever get write-access - it's not a fast process).
As a bonus, if you made your own PR, you'll get given a personal build of the plugin containing your changes, which can unblock you until we can get stuff merged - we both win ;-)

phucyall:
Wonderful. I did already create my own PR with just the fix for the exception (there is a workaround that doesn't require updated timeout). Looking forward to my own build of the plugin 😃

MarcelMue:
Just as a heads up: I am currently unable to implement any changes as I no longer have a setup to test on. The feedback from Eric came so much later, that I had given up on the hopes of ever getting something merged.

My workaround was to simply switch away from the plugin in the end.

phucyall:
If the development on this plugin picks up again (at least the merging starts happening) I would gladly re-do your changes with the feedback addressed. It would for sure be nice to be able to set the timeout right in the clone task. That's more of an enhancement though. The real bug was lack of null check before setting  "VSPHERE_IP" env. variable.

pjdarton:
@MarcelMue That's pretty much the situation that Jason (currently the only person with write-access) is in - he no longer has a setup to test on.  Also, the notifications (to Jason) about activity on the plugin stopped getting through, so he had no idea everyone was waiting for him.  He is now aware ;-)

@phucyall I'd suggest that you "subscribe" to the issues that I raised that aren't merged yet (71 to 78).  If you see a series of mergers happening there, that's your queue to get coding and submit a tidier PR that we can merge before the flurry of activity ceases ;-)

pjdarton:

I've not merged this one because the concerned raised earlier hadn't been addressed.
However this is now being handled by #79 (already merged) and #81 (in progress right now), so I'm going to close this one.

PR #67, opened 2017-02-21
Fix some minor findbugs warnings

mdkf:


pjdarton:

I've not merged #67 because I don't see these FindBugs warnings in either my IDE (I use Eclipse) or in the Jenkins build (findbugs doesn't seem to be being called) and I'm wary of making changes where I can't see the impact.
I would like us to have findbugs run over the code if we can, as code-analysis checking tools like this can find silly errors much more quickly than manual testing does, but it isn't something I know much about (I've never dealt with findbugs or telling maven to call it).
Also, the other changes made mean that this PR now has conflicts, so it's no longer a single button-push.

If you can walk me through how to verify these changes - where findbugs reports should show up, where the build should fail if there are issues found etc - and how to ensure that they're run every build, and update the PR so it can be merged, then I'll give it a shot.

pjdarton:
I've merged these changes, bit by bit, in #87 and #88.
The only bit I've not taken was updating the org.jenkins-ci.plugins/plugin version to 2.30 (as that forces us onto Java8 and I don't want that yet).

mdkf:
Thanks @pjdarton !

PR #69, opened 2017-03-01
Adding StatsD support

onlyvitriol:
Added baseline statsd support for the vSphere plugin in order to provide a deeper understanding of when automatically provisioned VMs succeed, and fail.

Currently supports two main metrics which follow the canonical logging namespace noticed in the original source:

vsphere-cloud.provision.failed
vsphere-cloud.provision.success
New dependency added on java-dogstatsd-client https://github.com/datadog/java-dogstatsd-client

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,6 +67,11 @@
             <type>jar</type>
         </dependency>
         <dependency>
+            <groupId>com.datadoghq</groupId>
+            <artifactId>java-dogstatsd-client</artifactId>
+            <version>2.3</version>
+        </dependency>
+        <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>credentials</artifactId>
             <version>1.28</version>
--- a/src/main/java/org/jenkinsci/plugins/vSphereCloud.java
+++ b/src/main/java/org/jenkinsci/plugins/vSphereCloud.java
@@ -57,7 +57,7 @@ public class vSphereCloud extends Cloud {
     @Deprecated
     private transient String password;
     private final int maxOnlineSlaves;
-    private final boolean useDDStatsd;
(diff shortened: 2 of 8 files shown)


pjdarton:

I've not merged this one because it adds a dependency, it now has conflicts and, personally, I think that the existing logging facilities should be enhanced to the point where they provide understanding of when VM provisioning fails rather than using yet-another-system for this.
I am, however, open to persuasion on this, e.g. if other plugins already do this.

So, to get this merged, you will have to:
1) Persuade me that this is a idea :-)   e.g. show that using statsd is "normal" for a cloud Jenkins plugin, e.g. by showing how it's used by similar plugins for similar purposes.
2) Update/Re-base this PR so that it can be merged.
3) Provide documentation (or links to documentation) that explain how to use this new functionality so we can copy/paste that into the plugin's wiki page.

onlyvitriol:
Due to the considerable delay in response, we have already created an internal fork.

pjdarton:
OK, understood.
Yeah, I agree, the delay in merging has been "considerable" - the (previous) plugin maintainer became unable to spend the time on the plugin (plus email changes meant he was no longer kept informed).
I can't guarantee I won't cause same situation in a year's time but, while I'm on this, I'll try to keep things moving ;-)

FYI there is a Jenkins process for taking over a plugin where the maintainer is unresponsive (which is how things are moving again) - if you get stuck on any other plugins (or this one), don't take "no response" for an answer and sit patiently - go grab the controls :-)

onlyvitriol:
Yeah it's good that there is a system in place for this :) Unfortunately my team wasn't in a position to take ownership of an open source project, so it wasn't exactly something we could have optioned on. 

Thanks for taking up ownership of this plugin though, and best of luck!

PR #74, opened 2017-06-28
Logging improvments.

pjdarton:
vSphereCloud's Log methods now log exceptions with Level.SEVERE instead of Level.INFO.  Non-exception logging is still Level.INFO.
vSphereCloudProvisionedSlave and vSphereCloudSlave's _terminate methods had some code paths that could fail silently - they now log something.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/vSphereCloudProvisionedSlave.java
+++ b/src/main/java/org/jenkinsci/plugins/vSphereCloudProvisionedSlave.java
@@ -57,13 +57,20 @@ public class vSphereCloudProvisionedSlave extends vSphereCloudSlave {
     @Override
     protected void _terminate(TaskListener listener) throws IOException, InterruptedException {
         super._terminate(listener);
-        final vSphereCloudLauncher launcher = (vSphereCloudLauncher) getLauncher();
-        if (launcher != null) {
-            final vSphereCloud cloud = launcher.findOurVsInstance();
-            if (cloud != null) {
+        try {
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/vSphereCloudProvisionedSlave.java
```diff
@@ -57,13 +57,20 @@ public vSphereCloudProvisionedSlave(String name, String nodeDescription,
     @Override
     protected void _terminate(TaskListener listener) throws IOException, InterruptedException {
         super._terminate(listener);
-        final vSphereCloudLauncher launcher = (vSphereCloudLauncher) getLauncher();
-        if (launcher != null) {
-            final vSphereCloud cloud = launcher.findOurVsInstance();
-            if (cloud != null) {
```
elordahl: Is there a reason this `cloud != null` check was removed?  Is an NPE possible here?

pjdarton: findOurVsInstance() will throw an exception - it won't return null.

pjdarton: I managed to miss off the extra Javadoc I'd added to make this clear from my PR.
I've added another commit that adds this.
(my local code had this javadoc already, but I'm in the process of un-picking my changes into a series of discrete PRs, and this comment got missed out.  Sorry for the confusion)

PR #78, opened 2017-06-28
Retention strategy

pjdarton:
The existing "keep until it's been idle for too long" strategy didn't appear to be working - VMs would get disposed of immediately after becoming idle.
This change replaces the implementation with one that follows the techniques used in the OpenStack plugin, as that one seems to work.
So, after this change, we now have a choice between a "use once" strategy (as before) or, now this works, we can keep VMs around until they've been unused for too long ("too long" is configurable).  This allows us to get a better build throughput as we do not have to wait for VMs to boot up for every build, as VMs can remain available for reuse after each build and only get killed off if they remain surplus to requirements for too long.


pjdarton:
This just does what #82 does, only #82 does it better.  I'll close this one and use #82 instead.

PR #80, opened 2017-07-25
[FIXED JENKINS-45786] Change idleMinutes to vsIdleMinutes

balleman:
This renames the VSphereCloudRetentionStrategy's idleMinutes field to vsIdleMinutes to prevent hiding the superclass field with the same name, which seems to have been preventing it from persisting properly across Jenkins restarts.  In my environment, this would lead to thrashing with short-lived jobs as nodes would go offline quickly instead of remaining available for reuse.  The workaround is to re-save the configuration via the web interface upon each Jenkins restart.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/vsphere/VSphereCloudRetentionStrategy.java
+++ b/src/main/java/org/jenkinsci/plugins/vsphere/VSphereCloudRetentionStrategy.java
@@ -10,16 +10,16 @@ import org.kohsuke.stapler.DataBoundConstructor;
 
 public class VSphereCloudRetentionStrategy extends CloudRetentionStrategy {
 
-    private final int idleMinutes;
+    private final int vsIdleMinutes;
 
     @DataBoundConstructor
-    public VSphereCloudRetentionStrategy(int idleMinutes) {
-        super(idleMinutes);
-        this.idleMinutes = idleMinutes;
+    public VSphereCloudRetentionStrategy(int vsIdleMinutes) {
+        super(vsIdleMinutes);
+        this.vsIdleMinutes = vsIdleMinutes;
     }
 
-    public int getIdleMinutes() {
-        return idleMinutes;
+    public int getVsIdleMinutes() {
+        return vsIdleMinutes;
     }
 
     @Override
--- a/src/main/resources/org/jenkinsci/plugins/vsphere/VSphereCloudRetentionStrategy/config.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/vsphere/VSphereCloudRetentionStrategy/config.jelly
@@ -1,5 +1,5 @@
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
-    <f:entry title="${%Idle Timeout}" field="idleMinutes">
+    <f:entry title="${%Idle Timeout}" field="vsIdleMinutes">
         <f:number default="5"/>
     </f:entry>
 </j:jelly>


pjdarton:
@balleman This change conflicts with my own (rather over-complicated) change #78.
I must confess that #78 is possibly overkill, so I would welcome your opinions / feedback on the subject.
I was under the impression that the vSphere cloud retention strategy didn't seem to work at all, which is why I created #78 which is to basically replace the vSphere implementation with one "inspired by" the OpenStack plugin's retention strategy.  It is possible that I was mistaken and that it was working perfectly but with the wrong timeout (as you identified in JENKINS-45786)...
...but it's also possible that I was not mistaken and that, while your change fixes the persistence issue, it still doesn't work.

Can you tell me what your own findings were?
If you can confirm that the retention strategy really does work (with such a simple fix of renaming the field) then that would obviously be a lot less risky than my "replace the entire class" PR.

...however, if we went with this PR, we would have to ensure that this seamlessly "upgraded" from the old version - I believe that there's some tricks one can pull with the data-bound constructors to ensure backwards-compatibility...

balleman:
@pjdarton With this change, or with re-saving the configuration after a Jenkins restart, the vSphere cloud retention strategy was working as expected for me, with nodes terminating after they were idle for the configured interval.  The symptom you describe in #78 is consistent with what I would see after a restart without re-saving the configuration - the node was disposed immediately (well, on the first retention strategy check), without waiting for the configured timeout.  

I agree there is a seamless upgrade problem with changing the field name--the previously configured value is lost with this change, though I've not researched how to address it.

pjdarton:
@balleman OK, it sounds like you've found a *far* more elegant solution to the problem I encountered than the "solution" I came up with :-)

If I can figure out how to deal with the upgrade problem and suggest code changes, are you willing to modify your pull-request and test the results?

balleman:
@pjdarton Yes, happy to accept changes and test.

pjdarton:
Hmm, this backwards-compatibility stuff is non-trivial...

My understanding (based on some googling + earlier experience) is that the DataBoundConstructor gets called by the Jelly code when the user presses "save", but it isn't called (at all!) when the class gets de-serialized from its XML form.  When the XML config is loaded, the class gets created and its fields set by reflection, and then the readResolve() method is called (if the class has one).
So we'll need a "private transient Integer idleMinutes" field (transient because we don't want it persisted when our data gets saved, but present because we want the de-serialization to have somewhere to store the data), which will get populated when the class gets de-serialised from its XML form, and then we'll need a private Object readResolve() method that goes something like { if (idleMinutes!=null) { return new VSphereCloudRetentionStrategy(idleMinutes.intValue()); } else { return this; } }

I found https://wiki.jenkins.io/display/JENKINS/Hint+on+retaining+backward+compatibility which explains it fairly well, although it doesn't really explain why we're seeing this bug in the first place...

Disclaimer: As this bug is all about the data failing to be respected when the class is de-serialized, it's quite possible that the idleMinutes field won't be set anyway (precisely because of this bug, e.g. if the de-serialization code isn't serialising/de-serializing the field in the parent class).  However it's also equally possible that merely adding the readResolve() method will sort the original bug without renaming the field to vsIdleMinutes.
Some experimentation may be required...
I would suggest that you take a look at the XML that Jenkins generated and see if the value entered was written into it in the first place - if the XML on disk doesn't contain the value then the bug we're solving isn't that "it doesn't get read" it's more than "it didn't get written".  If the data isn't on disk to start with then there's no data to be backwards-compatible with and the existing code can be used as-is (the damage is already done!).

balleman:
The config is being saved into the XML (with an unchanged plugin):
          <retentionStrategy class="org.jenkinsci.plugins.vsphere.VSphereCloudRetentionStrategy">
            <idleMinutes>6</idleMinutes>
            <idleMinutes defined-in="org.jenkinsci.plugins.vsphere.VSphereCloudRetentionStrategy">6</idleMinutes>
          </retentionStrategy>

I've tried adding a readResolve() that just calls the constructor, and it appears to work.  The idleMinutes is being set on the VSphereCloudRentionStrategy instance, just not the field with the same name in the superclass.

    private Object readResolve() {
        // without this, super.idleMinutes is not restored from persistence
        return new VSphereCloudRetentionStrategy(idleMinutes);
    }

Provided there is nothing wrong with this approach, would you prefer this approach committed in this PR, or to open a different one?  I'm leaning towards a new one since none of the changes in the existing branch would be retained.

pjdarton:
As pull requests are "free", you might as well create a new PR.

balleman:
Closed in favor of #82

PR #81, opened 2017-07-28
Minor fixes for adding timeOutInSeconds to Clone and Deploy by MarcelMue

phucyall:
I ported the changes created by MarcelMue into a new PR with the code review feedback addressed.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/vsphere/builders/Clone.java
+++ b/src/main/java/org/jenkinsci/plugins/vsphere/builders/Clone.java
@@ -42,7 +42,7 @@ import com.vmware.vim25.mo.VirtualMachineSnapshot;
 
 public class Clone extends VSphereBuildStep {
 
-	private final int TIMEOUT_DEFAULT = 60;
+    private final Integer TIMEOUT_DEFAULT = 60;
 
     private final String sourceName;
     private final String clone;
@@ -53,13 +53,13 @@ public class Clone extends VSphereBuildStep {
     private final String folder;
     private final String customizationSpec;
     private final boolean powerOn;
-    private final int timeoutInSeconds;
+    private final Integer timeoutInSeconds;
     private String IP;
 
     @DataBoundConstructor
     public Clone(String sourceName, String clone, boolean linkedClone,
                  String resourcePool, String cluster, String datastore, String folder,
-                 boolean powerOn, int timeoutInSeconds, String customizationSpec) throws VSphereException {
+                 boolean powerOn, Integer timeoutInSeconds, String customizationSpec) throws VSphereException {
         this.sourceName = sourceName;
         this.clone = clone;
         this.linkedClone = linkedClone;
@@ -70,10 +70,10 @@ public class Clone extends VSphereBuildStep {
         this.customizationSpec=customizationSpec;
         this.powerOn=powerOn;
         if (timeoutInSeconds != null) {
-        	this.timeoutInSeconds = timeoutInSeconds;
+            this.timeoutInSeconds = timeoutInSeconds;
         }
         else {
-        	this.timeoutInSeconds = TIMEOUT_DEFAULT;
+            this.timeoutInSeconds = TIMEOUT_DEFAULT;
         }
     }
 
@@ -113,9 +113,9 @@ public class Clone extends VSphereBuildStep {
         return powerOn;
     }
     
-	public int getTimeoutInSeconds() {
-		return timeoutInSeconds;
-	}
+    public Integer getTimeoutInSeconds() {
+        return timeoutInSeconds;
+    }
 
     @Override
     public void perform(@Nonnull Run<?, ?> run, @Nonnull FilePath filePath, @Nonnull Launcher launcher, @Nonnull TaskListener listener) throws InterruptedException, IOException {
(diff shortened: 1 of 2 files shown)


phucyall:
This is a clone of https://github.com/jenkinsci/vsphere-cloud-plugin/pull/65 with feedback addressed

src/main/java/org/jenkinsci/plugins/vsphere/builders/Clone.java
```diff
@@ -51,12 +53,13 @@
...
     private final String customizationSpec;
     private final boolean powerOn;
+    private final int timeoutInSeconds;
     private String IP;
 
     @DataBoundConstructor
     public Clone(String sourceName, String clone, boolean linkedClone,
                  String resourcePool, String cluster, String datastore, String folder,
-                 boolean powerOn, String customizationSpec) throws VSphereException {
+                 boolean powerOn, int timeoutInSeconds, String customizationSpec) throws VSphereException {
```
pjdarton: "int" should be "Integer" in order to cope with old configuration data where this parameter did not exist.

src/main/java/org/jenkinsci/plugins/vsphere/builders/Clone.java
```diff
@@ -42,6 +42,8 @@
 
 public class Clone extends VSphereBuildStep {
 
+	private final int TIMEOUT_DEFAULT = 60;
```
pjdarton: Indentation incorrect - use spaces, not tabs (tabs are evil - they never display the same as some people set them to 4, some 2, some 8 etc).

src/main/java/org/jenkinsci/plugins/vsphere/builders/Clone.java
```diff
@@ -66,6 +69,12 @@ public Clone(String sourceName, String clone, boolean linkedClone,
         this.folder=folder;
         this.customizationSpec=customizationSpec;
         this.powerOn=powerOn;
+        if (timeoutInSeconds != null) {
+        	this.timeoutInSeconds = timeoutInSeconds;
```
pjdarton: Once the timeoutInSeconds is an Integer, you should use ... = timeoutInSeconds.intValue(); to explicitly un-box the value.

src/main/java/org/jenkinsci/plugins/vsphere/builders/Clone.java
```diff
@@ -42,6 +42,8 @@
 
 public class Clone extends VSphereBuildStep {
 
+    private final Integer TIMEOUT_DEFAULT = 60;
```
pjdarton: In this case, Integer should be int - this one isn't nullable.

(Sorry, I should have spotted this earlier)

src/main/java/org/jenkinsci/plugins/vsphere/builders/Clone.java
```diff
@@ -51,12 +53,13 @@
     private final String folder;
     private final String customizationSpec;
     private final boolean powerOn;
+    private final Integer timeoutInSeconds;
```
pjdarton: I think that, unless we want to store a null/empty timeoutInSeconds field, this should be "int".
Pros/Cons...
If we have this field as an "int" then, if the user opens and presses "Save" on the config without setting the value, this field will be added with the default value.
If we have this field as an "Integer" then the field can be persisted as having no value set, meaning that the persisted data is basically saying "we want the default timeout to be used, whatever that is".  In this case, the code should be modified to do the "if this is null, use the default value" at execution-time instead of at construction time.

Personally I think it'd be simpler to have this as an int, and have the get method also return an int.
(if you want it as an Integer then you'll have to delay the "if null..." stuff until the perform(...) method.

phucyall: Yeah sorry, I'm pretty new to all this code so bear with me a bit.

I like the idea of keeping everything except the constructor as int so I'll do that.

pjdarton: I'm only just *one* step ahead of you at present - it's a "learning experience" for both of us :-)
...unfortunately, this means that I'm sometimes pointing you in the wrong direction :-(

I've just been reading https://wiki.jenkins.io/display/JENKINS/Hint+on+retaining+backward+compatibility and that's reminded me that, when Jenkins loads the configuration off disk, it doesn't actually call the constructor at all.
This means that we can't rely on the constructor setting the field to a value before perform() gets run because, where a user has an old job (i.e. no timeoutInSeconds field), and their job runs with the new plugin code (which is looking for a timeoutInSeconds field), then perform() will find that field is set to zero.
So I think that we need to decide what we want the default to be (should it wait forever, or one minute, or what?  e.g. In my experience, one minute isn't enough for a Windows VM!), and then defer deciding what to until within the perform() method (which isn't what I'd do if I could rely on the constructor being called!).

I'm not familiar with what this code did before there was a timeout - did it wait "forever"?  If so, I'd suggest that this is what the default behavior should be as, that way, existing customers will see no difference, but can decide to take advantage of the new flexibility if they choose to.

(I do apologize - I was expecting this part to be *way* simpler than it's turning out to be!)

PR #92, opened 2018-04-24
Don't force the use of underscore to separate slave prefix from number.

joshk0:
* Underscores complicate things when you want to create DNS records for
  the ephemeral slaves. Underscores are not valid characters for DNS :(
* The user can always add the underscore back in their config.

This should be documented as a breaking functional change but I don't
really see that people would be designing tons of functionality around
this underscore. As a result, there is no migration from pre- to post-
removal versions of the plugin.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/vsphere/tools/CloudProvisioningAlgorithm.java
+++ b/src/main/java/org/jenkinsci/plugins/vsphere/tools/CloudProvisioningAlgorithm.java
@@ -70,7 +70,7 @@ public final class CloudProvisioningAlgorithm {
         final int maxAttempts = hasCap ? (templateInstanceCap) : 100;
         for (int attempt = 0; attempt < maxAttempts; attempt++) {
             final String suffix = hasCap ? calcSequentialSuffix(attempt) : calcRandomSuffix(attempt);
-            final String nodeName = cloneNamePrefix + "_" + suffix;
+            final String nodeName = cloneNamePrefix + suffix;
             if (!existingNames.contains(nodeName)) {
                 return nodeName;
             }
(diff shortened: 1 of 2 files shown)


pjdarton:
Yes, it's on my list of things to do.  I've been out of the office for a week (and was busy the previous week) but, yes, it's on the list - I'm delaying, not ignoring you entirely ;-)

pjdarton:
Hi.  Sorry for the delay - I had a load of issues with the docker-plugin I needed to fix - but I'm back on the case now and I've had a look at this.
I have good news and I have bad news.

The bad news is that I can't merge this as-is.
I can see that the code changes do achieve exactly the effect you were after (which is good), but this PR doesn't include all the other bits that are needed when making changes that affect the configuration / the way the configuration is used.
e.g. if I merged this & released a new version of the plugin, and then someone upgraded their plugin, they'd "suddenly" find that their node names were missing the underscore, and that might be a breaking change for some users.
We have to ensure that, when upgrading the plugin, "nothing changes" by default.

The good news is that the extra changes required aren't too arduous and so I've come up with #93 which includes additional changes that address my concerns.  I've (very briefly) tried it out to verify that the configuration upgrade code works as I expect, but it _will need testing properly before release_ and I'd like to enlist your help on that front.

FYI the extra changes ensure that, if someone upgrades from an earlier version of the plugin then:
 - the underscore is no longer hard-coded but now appears in the template configuration.
 - that underscore can be removed from the template configuration.
 - the online help no longer says that there's always an underscore added.

Finally, ***can you please test*** #93 and let me know what you think.
You'll find a .hpi file to download from the [artifacts section](https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fvsphere-cloud-plugin/detail/PR-93/1/artifacts) on the ci.jkenkins.io site that's linked to from the PR's "all checks" section.

pjdarton:
Superseded by #93 
All changes in here are in there.
Closing this one.  All further discussion to be done in #93 instead.

PR #93, opened 2018-05-14
Stop hard-coding underscore in clone names

pjdarton:
#92 made changes to the functionality to remove the hard-coded underscore between a template's cloneNamePrefix and the random ID in the slave node names.
However, these changes didn't update the online help text, the javadoc, or handle the upgrade path for existing users (who would "suddenly" find their clone names were lacking the underscore after an upgrade).
This PR incorporates #92 but also includes changes to address the above.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
+++ b/src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
@@ -355,7 +355,7 @@ public class vSphereCloudSlaveTemplate implements Describable<vSphereCloudSlaveT
         // Earlier versions of the code may have stored configuration in a
         // different way but in the same kind of fields, so we need an explicit
         // versioning to know how to mutate the data.
-        if( configVersion<=0) {
+        if (configVersion <= 0) {
             LOGGER.log(Level.CONFIG,
                     "{0} loaded old configuration that had hard-coded underscore at the end of the cloneNamePrefix.",
                     this);
@@ -373,7 +373,7 @@ public class vSphereCloudSlaveTemplate implements Describable<vSphereCloudSlaveT
                     + CURRENT_CONFIG_VERSION
                     + ".  Either CURRENT_CONFIG_VERSION is incorrect or the readResolve method is not setting configVersion when it upgrades the data.");
         }
-        if( configVersion > CURRENT_CONFIG_VERSION ) {
+        if (configVersion > CURRENT_CONFIG_VERSION) {
             LOGGER.log(Level.WARNING,
                     "{0} was defined by a later version of the plugin "
                             + "(one that saved with configVersion={1}, whereas this version of the plugin is expecting {2}).  "
--- a/src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
+++ b/src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
@@ -82,7 +82,9 @@ public class vSphereCloudSlaveTemplate implements Describable<vSphereCloudSlaveT
     protected static final SchemeRequirement HTTP_SCHEME = new SchemeRequirement("http");
     protected static final SchemeRequirement HTTPS_SCHEME = new SchemeRequirement("https");
 
-    private final String cloneNamePrefix;
+    private int configVersion;
+    private static final int CURRENT_CONFIG_VERSION = 1;
+    private String cloneNamePrefix; // almost final
     private final String masterImageName;
     private Boolean useSnapshot; // almost final
     private final String snapshotName;
@@ -148,6 +150,7 @@ public class vSphereCloudSlaveTemplate implements Describable<vSphereCloudSlaveT
                                      final RetentionStrategy<?> retentionStrategy,
                                      final List<? extends NodeProperty<?>> nodeProperties,
                                      final List<? extends VSphereGuestInfoProperty> guestInfoProperties) {
+        this.configVersion = CURRENT_CONFIG_VERSION;
         this.cloneNamePrefix = cloneNamePrefix;
         this.masterImageName = masterImageName;
         this.snapshotName = snapshotName;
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
```diff
@@ -349,6 +352,38 @@ protected Object readResolve() {
                 LOGGER.log(Level.CONFIG, " - Failed to reconfigure strategy", ex);
             }
         }
+        // Earlier versions of the code may have stored configuration in a
+        // different way but in the same kind of fields, so we need an explicit
+        // versioning to know how to mutate the data.
+        if( configVersion<=0) {
```
joshk0: nit: inconsistent parens (looks like this project generally goes for the standard-ish `if (...) {` with space after the if and space after the close paren.)

src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
```diff
@@ -349,6 +352,38 @@ protected Object readResolve() {
...
+            LOGGER.log(Level.CONFIG,
+                    "{0} loaded old configuration that had hard-coded underscore at the end of the cloneNamePrefix.",
+                    this);
+            // In version one, the underscore was removed from the code so we
+            // have to put it in the data (the user is then free to change it to
+            // something else if they want to).
+            this.cloneNamePrefix = this.cloneNamePrefix + "_";
```
joshk0: Can `this.cloneNamePrefix` be empty going into this method, e.g. perhaps if there was no config object at all to begin with? Is `readResolve()` only called if there is an existing config, and if so, is `cloneNamePrefix` guaranteed to be non-null and non-empty?

src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
```diff
@@ -349,6 +352,38 @@ protected Object readResolve() {
...
+        }
+        // Note: Subsequent changes dependent on configVersion should go above
+        // this line.
+        if (configVersion < CURRENT_CONFIG_VERSION) {
+            throw new IllegalStateException("Internal error: configVersion==" + configVersion
+                    + " at end of readResolve method, but the current config version should be "
+                    + CURRENT_CONFIG_VERSION
+                    + ".  Either CURRENT_CONFIG_VERSION is incorrect or the readResolve method is not setting configVersion when it upgrades the data.");
+        }
+        if( configVersion > CURRENT_CONFIG_VERSION ) {
```
joshk0: nit: paren spacing

src/main/resources/org/jenkinsci/plugins/vSphereCloudSlaveTemplate/help-cloneNamePrefix.html
```diff
@@ -1,6 +1,6 @@
 <div>
 Controls the name by which the virtual machine will be known to vSphere.
-The plug-in will append an underscore and a unique ID to this prefix in order to generate unique names.<p>
+The plug-in will append a unique ID to this prefix in order to generate unique names.<p>
```
joshk0: HTML nit (maybe applies throughout) -- do this

```
<p>stuff</p>

<p>more stuff</p>
```

src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
pjdarton: There are no guarantees.
 - If it was empty then the slave node names would previously have been named "_someRandomId" and this code will ensure that this remains the case.
 - If it was null then the slave node names would previously have been named "null_someRandomId" and this code will ensure that this remains the case.
i.e. The old behavior is maintained, but a bad config won't be magically made good.

In an ideal world, we'd have some sort of "post deserialization fix-up" at the cloud level that ensured that, within the vSphereCloud, the prefix was non-null, non-empty and unique for each template, but I'm not 100% sure where that code would have to go / get-called-from in order to make it work in all situations.

src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
pjdarton: I'm quite happy with nit-picking comments; I'll sort those out - I dislike messy/inconsistent code too.

src/main/resources/org/jenkinsci/plugins/vSphereCloudSlaveTemplate/help-cloneNamePrefix.html
pjdarton: Personally, I'm not much of a fan of putting formatting/layout elements at the start/end of lines that contain content.  I prefer to keep layout and content on their own lines. e.g. 
```html
<p>
stuff
</p>
<p>
more stuff
</p>
```

However, I did try wrapping the first line of text (line 2 in the file) in `<p>` ... `</p>` and found that this then added an unsightly amount of vertical whitespace in the help text (between the top border of the help box and the top of the first line of text).
There's a border around the existing help-text that's about as wide & tall as an "M" character, and the word "Controls" at the start and the "(from vSphere Plugin)" text at the end are both indented by equal amounts.
If we make the initial text a paragraph then the layout becomes unbalanced and there's "two Ms" worth of vertical whitespace before the first word "Controls".
i.e. it might be "more correct" HTML but it's also more ugly.

I suspect that the HTML generated by Jenkins to wrap the help text was designed more for "simple text" than HTML containing lots of layout formatting.  I suspect that it was designed to make things "look right" if they start and end with no vertical whitespace.

I'm not an HTML expert, so I did `mvn hpi:run` and did some experimentation with the layout ... and I found that if we remove `<p>` and `</p>` completely and instead separate the initial text (lines 2 and 3 in this file) from the "Note" (line 4) by `<br><br>` then that looks good too (in my opinion), and it also uses half an M less vertical whitespace between the "Note" text and the "(from vSphere Plugin)" text, which appeals to my sense of neatness.

...also (a brief google later), doing `<p style="margin-top:0;">stuff</p>` for the first makes it look acceptable when rendered too, but I suspect that messing with styling here would be considered worse than just using `<br><br>` would be.

I'm open to ideas but unless we can settle on something that's both "good HTML practice" and also "looks correct when rendered" then I think it may be best to leave that code as-is.

pjdarton:
@joshk0 A few days ago you said you were going to test this ...  What's the verdict?

pjdarton:
@joshk0 Nudge.
I've been running this locally for a while now and it doesn't seem to break anything.
If there's any further changes you think are necessary, please let me know ASAP, otherwise I'll likely push out a release with these changes as-is.
(if you let me know that you're happy with the changes as-is, I may push out a release sooner than I otherwise might)

joshk0:
I'm sorry Peter, I wasn't able to test this. However I really don't see any downside, I was happy with it as is and you've made it even safer and conservative so i feel like I would be happy to see this in a release. Thanks for following up.

pjdarton:
@joshk0 FYI release 2.18 was uploaded a few minutes ago; it should get indexed in the next 12 hours or so and thus appear as an available update in your Jenkins server(s) within 48 hours.
These code changes were included so you should be safe to update and use the official release instead of a private build.

PR #96, opened 2018-09-20
Request for feedback: Add new fixed-lifespan cloud retention strategy.

rhencke:
We have found a need for an alternate cloud retention strategy for our builds.

The keep-until-idle strategy is very good, but the issue we hit is that our builds
are continuously running so agent never go idle.  Eventually, the agents will hit
some issue we must address, such as running out of disk space.  While we could
work to fix the agents, it's much easier to just delete them and let them be recreated.

So, we have introduced a fixed-life strategy to the vSphere plugin.  
The fixed-lifespan strategy provisions an agent for a fixed amount of time.
After the amount of time has expired, the agent is disconnected, any builds
finished, then the agent is removed.  This prevents incurring the costs of the
build-once strategy, which involves bootup/provisioning/no cache costs, but
ensures the agents are recycled on a regular basis which helps prevent issues.

While, currently, we do this as a separate retention strategy, I could also see this
simply being an option that could be added to the existing keep-until-idle strategy.

I was hoping to get your thoughts, and if interested, submit this for inclusion.

Thank you.


src/main/java/org/jenkinsci/plugins/vsphere/FixedLifespanCloudRetentionStrategy.java
```diff
@@ -0,0 +1,114 @@
...
+        final long ageMillis = System.currentTimeMillis() - creationTimeMillis;
+        final String cname = c.getName();
+        if (!isAtEndOfLife() && ageMillis > TimeUnit2.MINUTES.toMillis(lifespanMinutes)) {
+            LOGGER.log(Level.FINE, "Will terminate {0} once idle - lifespan of {1} minutes reached.", new Object[] { cname, lifespanMinutes });
+            setAtEndOfLife();
+            final VSphereOfflineCause cause = new VSphereOfflineCause(Messages._fixedLifespanCloudRetentionStrategy_OfflineReason_LifespanReached(String.valueOf(lifespanMinutes)));
+            try {
+                c.disconnect(cause).get();
+            } catch (InterruptedException | ExecutionException e) {
+                LOGGER.log(WARNING, "Failed to disconnect " + cname, e);
```
pjdarton: If this fails to disconnect, it looks like `setAtEndOfLife()` will still be set so that no further attempts will be made to disconnect the slave (until Jenkins gets restarted).
That doesn't seem right to me.

src/main/java/org/jenkinsci/plugins/vsphere/FixedLifespanCloudRetentionStrategy.java
```diff
@@ -0,0 +1,114 @@
...
+            if (computerNode != null) {
+                try {
+                    LOGGER.log(Level.FINER, "Initiating termination of {0}.", cname);
+                    computerNode.terminate();
+                } catch (InterruptedException | IOException e) {
+                    LOGGER.log(WARNING, "Failed to terminate " + cname, e);
+                }
+            }
+        }
+        return 1; // re-check in 1 minute
```
pjdarton: Once a machine has reached end-of-life then re-checking every minute is good.
However, before it's reached end-of-life, we don't need to check every minute - we should return "number of minutes until we reach end of line".

src/main/java/org/jenkinsci/plugins/vsphere/FixedLifespanCloudRetentionStrategy.java
```diff
@@ -0,0 +1,114 @@
...
+        super.start(c);
+        c.connect(false);
+    }
+
+    @Override
+    public boolean isAcceptingTasks(AbstractCloudComputer c) {
+        return !isAtEndOfLife();
+    }
+
+    private transient boolean atEndOfLife;
```
pjdarton: I'm not a big fan of having class fields scattered throughout the code - I much prefer to see all the fields at the top.

src/main/java/org/jenkinsci/plugins/vsphere/FixedLifespanCloudRetentionStrategy.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private synchronized boolean isAtEndOfLife() {
```
pjdarton: I don't think that `synchronized` _here_ is going to achieve thread-safety.
e.g. the `check` method calls `isAtEndOfLife()` twice, so if you only lock in here, `check` might get two different values when it calls the method if another thread has called `setAtEndOfLife()` in between.
You probably want to make `check()` be `synchronized` (or add a `synchronized` section within it) so that you can ensure that you get a consistent and controlled logic flow within that method.
...also, if you put it there, you can reduce the number of synchronized blocks to just one - thread synchronization is expensive so we don't want to do that any more than we absolutely have to.
...and I'd suggest that you check with other retention strategies (in this and other plugins) and verify that it's needed at all because, if we don't need to make this code thread-safe, we shouldn't add code making it look like it is.

src/main/resources/org/jenkinsci/plugins/vsphere/FixedLifespanCloudRetentionStrategy/config.jelly
```diff
@@ -0,0 +1,6 @@
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
+    <f:entry title="${%Lifespan Timeout}" field="lifespanMinutes">
+        <f:number default="60"/>
+    </f:entry>
+</j:jelly>
```
pjdarton: Please add a linefeed to the end of the file.

src/main/resources/org/jenkinsci/plugins/vsphere/Messages.properties
```diff
@@ -1 +1,2 @@
 runOnceCloudRetentionStrategy.OfflineReason.BuildHasRun=VSphere Cloud Slave configured was to run one build only
+fixedLifespanCloudRetentionStrategy.OfflineReason.LifespanReached=VSphere Cloud Agent has reached the end of its configured lifespan of {0} minutes
```
pjdarton: Please add a linefeed to the end of the file.

pjdarton:
This PR has been inactive for some time now...
@rhencke are you still interested in completing this and getting it merged in?

I'm not keen on having PRs open for ages - I'd prefer that we either work on them until they're complete and get them merged in, or we close them if that's not going to happen.

PR #98, opened 2018-10-04
[WIP] Add VM creation/destruction hook points.

pjdarton:
Can now get the master to execute arbitrary code just after a VM is created and just before it is destroyed.
That means an admin could write a script to grab important data off the VM before it's destroyed.

We've got issues that are proving difficult to diagnose, and this facility is an attempt at making diagnosis easier.  It it proves useful enough to warrant keeping, I'll ensure it's production-strength; as present, it's intended for non-production usage and debugging only.


pjdarton:
I'm never going to merge this; it's low-quality code that I wrote in order to do some debugging of my work vSphere infrastructure and I have no inclination to bring it to a point where it could be merged as it served its purpose.

PR #99, opened 2018-10-24
Fix button text in folder configuration

albers:
On the folder configuration page, the button for adding a cloud is wrongly labelled _Add vSphere Template_, see

add-vsphere-template

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/folder/FolderVSphereCloudProperty/config.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/folder/FolderVSphereCloudProperty/config.jelly
@@ -2,7 +2,7 @@
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:f="/lib/form">
     <f:section title="${%vSphere Cloud}">
         <f:entry title="${%Clouds}" description="${%List of Clouds}">
-            <f:repeatable field="clouds" add="${%Add vSphere Template}">
+            <f:repeatable field="clouds" add="${%Add vSphere Cloud}">
                 <st:include page="config-inner.jelly" class="${descriptor.clazz}"/>
                 <div align="right">
                     <f:repeatableDeleteButton value="${%Delete vSphere Template}"/>


albers:
@pjdarton Does a folder-level cloud & template configuration work for you?

If I define both at this level, the node labels associated with the templates are not recognized.
The job configurations do no even show a "Restrict where this project can be run" checkbox.

I have to define the cloud and templates in the global configuration.

pjdarton:
I don't use folder-level configuration at all where I work; I only use global configuration.

I was not involved when this "Folder" functionality was added and cannot comment on what features or functionality it was intended to add, and hence have no idea if it solved the problems that the author (of that feature) intended it to solve.
My (limited) understanding of the [folder plugin](https://plugins.jenkins.io/cloudbees-folder) is that it's of limited use on its own and is only fully functional as the (non-free) [folders plus](https://go.cloudbees.com/docs/cloudbees-documentation/cje-user-guide/index.html#folders-plus) plugin.  As I only use the free Jenkins stuff at work, I've never fully explored what problems the non-free functionality is meant to solve.
My assumption was that the support for folders that was added to this plugin would be useful with both the free and the non-free folder code, but written in a way that did not prevent the plugin from working without either.

All I know for sure is that it didn't break everything else :-)

If your jobs are missing a "Restrict where this project can be run" checkbox then that's a separate issue; I do not believe that it'll have anything to do with this plugin (maybe it might be related to how you're using the folders plugin, or maybe something else).
I'd suggest that you try defining a normal "general purpose" job, both at the top level and within folders, and setting "Restrict where this project can be run" in there.  I'd also suggest that you enable lots of logging and see what Jenkins is asking the vSphere plugin to do when you ask it to run a build.

albers:
One of the features of the *folders plus* plugin is "restrict the use of slaves to specific folders". Probably this also affects the cloud plugins and therefore vSphere when used with the basic *folders* plugin.

> If your jobs are missing a "Restrict where this project can be run" checkbox then that's a separate issue; I do not believe that it'll have anything to do with this plugin 

Assuming that per folder cloud configuration only works with the *folders plus* plugin, this is expected behaviour: The checkbox is hidden unless agents are configured. If the folder level cloud configuration is ignored, no agents are configured and no restriction on where to run the job can be applied.

PR #100, opened 2018-12-11
[JENKINS-60029] - mark VSphereCloudRetentionStrategy as an extension

casz:
fixes [JENKINS-60029](https://issues.jenkins-ci.org/browse/JENKINS-60029)

This should fix vSphere cloud retention strategy not being configurable from JCasc.

I might be interested in giving the plugin a little love.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -98,6 +98,19 @@
             <artifactId>workflow-basic-steps</artifactId>
             <version>2.0</version>
         </dependency>
+        <dependency>
+            <groupId>io.jenkins</groupId>
+            <artifactId>configuration-as-code</artifactId>
+            <version>1.32</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.jenkins</groupId>
+            <artifactId>configuration-as-code</artifactId>
+            <version>1.32</version>
+            <classifier>tests</classifier>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <repositories>
--- a/src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
+++ b/src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
@@ -182,8 +182,8 @@ public class vSphereCloudSlaveTemplate implements Describable<vSphereCloudSlaveT
         this.targetResourcePool = targetResourcePool;
         this.targetHost = targetHost;
         this.credentialsId = credentialsId;
-        this.nodeProperties = nodeProperties;
-        this.guestInfoProperties = guestInfoProperties;
+        this.nodeProperties = Util.fixNull(nodeProperties);
+        this.guestInfoProperties = Util.fixNull(guestInfoProperties);
         this.launcher = launcher;
         this.retentionStrategy = retentionStrategy;
         readResolve();
--- a/src/main/java/org/jenkinsci/plugins/vsphere/RunOnceCloudRetentionStrategy.java
+++ b/src/main/java/org/jenkinsci/plugins/vsphere/RunOnceCloudRetentionStrategy.java
@@ -16,6 +16,7 @@
 
 package org.jenkinsci.plugins.vsphere;
 
+import hudson.Extension;
 import hudson.model.ExecutorListener;
 import hudson.model.Descriptor;
 import hudson.model.Executor;
@@ -168,6 +169,7 @@ public class RunOnceCloudRetentionStrategy extends CloudRetentionStrategy implem
     }
 
     @Restricted(NoExternalUse.class)
+    @Extension
     public static final DescriptorImpl DESCRIPTOR = new DescriptorImpl();
 
     public static final class DescriptorImpl extends Descriptor<RetentionStrategy<?>> {
--- a/src/main/java/org/jenkinsci/plugins/vsphere/VSphereCloudRetentionStrategy.java
+++ b/src/main/java/org/jenkinsci/plugins/vsphere/VSphereCloudRetentionStrategy.java
@@ -1,5 +1,6 @@
 package org.jenkinsci.plugins.vsphere;
 
+import hudson.Extension;
 import hudson.model.Descriptor;
 import hudson.slaves.CloudRetentionStrategy;
 import hudson.slaves.RetentionStrategy;
@@ -28,6 +29,7 @@ public class VSphereCloudRetentionStrategy extends CloudRetentionStrategy {
     }
 
     @Restricted(NoExternalUse.class)
+    @Extension
     public static final DescriptorImpl DESCRIPTOR = new DescriptorImpl();
 
     public static final class DescriptorImpl extends Descriptor<RetentionStrategy<?>> {
--- a/src/test/java/org/jenkinsci/plugins/vsphere/tools/ConfigurationAsCodeTest.java
+++ b/src/test/java/org/jenkinsci/plugins/vsphere/tools/ConfigurationAsCodeTest.java
@@ -1,0 +1,94 @@
+package org.jenkinsci.plugins.vsphere.tools;
+
+import hudson.model.Node.Mode;
+import hudson.slaves.ComputerLauncher;
+import hudson.slaves.JNLPLauncher;
+import hudson.slaves.RetentionStrategy;
+import io.jenkins.plugins.casc.ConfigurationContext;
+import io.jenkins.plugins.casc.ConfiguratorRegistry;
+import io.jenkins.plugins.casc.misc.ConfiguredWithCode;
+import io.jenkins.plugins.casc.misc.JenkinsConfiguredWithCodeRule;
+import io.jenkins.plugins.casc.model.CNode;
+import java.util.List;
+import org.jenkinsci.plugins.vSphereCloud;
+import org.jenkinsci.plugins.vSphereCloudSlaveTemplate;
+import org.jenkinsci.plugins.vsphere.RunOnceCloudRetentionStrategy;
+import org.jenkinsci.plugins.vsphere.VSphereGuestInfoProperty;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+import static io.jenkins.plugins.casc.misc.Util.getJenkinsRoot;
+import static io.jenkins.plugins.casc.misc.Util.toStringFromYamlFile;
+import static io.jenkins.plugins.casc.misc.Util.toYamlString;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasSize;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.core.Is.is;
+
+public class ConfigurationAsCodeTest {
+
+    @ClassRule
+    @ConfiguredWithCode("configuration-as-code.yml")
+    public static JenkinsConfiguredWithCodeRule r = new JenkinsConfiguredWithCodeRule();
+
+    @Test
+    public void should_support_configuration_as_code() {
+        vSphereCloud cloud = (vSphereCloud) r.jenkins.clouds.get(0);
+        assertThat(cloud.getVsDescription(), is("Company vSphere"));
+        assertThat(cloud.getVsHost(), is("https://company-vsphere"));
+        assertThat(cloud.getInstanceCap(), is(100));
+        assertThat(cloud.getMaxOnlineSlaves(), is(0));
+        List<? extends vSphereCloudSlaveTemplate> templates = cloud.getTemplates();
+        assertThat(templates, hasSize(1));
+        vSphereCloudSlaveTemplate template = templates.get(0);
+        assertThat(template, notNullValue());
+        assertThat(template.getCloneNamePrefix(), is("windows-"));
+        assertThat(template.getCluster(), is("Company"));
+        assertThat(template.getDatastore(), is("Company-FMD-01"));
+
+        assertThat(template.getForceVMLaunch(), is(true));
+        assertThat(template.getLabelString(), is("windows vsphere"));
+        assertThat(template.getLaunchDelay(), is(60));
+        assertThat(template.getLimitedRunCount(), is(1));
+        assertThat(template.getLinkedClone(), is(true));
+        assertThat(template.getMasterImageName(), is("windows-server-2019"));
+        assertThat(template.getMode(), is(Mode.EXCLUSIVE));
+        assertThat(template.getNumberOfExecutors(), is(1));
+        assertThat(template.getRemoteFS(), is("C:/jenkins"));
+        assertThat(template.getResourcePool(), is("Resources"));
+        assertThat(template.getSaveFailure(), is(false));
+        assertThat(template.getTemplateInstanceCap(), is(5));
+        assertThat(template.getUseSnapshot(), is(true));
+        assertThat(template.getWaitForVMTools(), is(true));
+        List<? extends VSphereGuestInfoProperty> guestInfoProperties = template.getGuestInfoProperties();
+        assertThat(guestInfoProperties, hasSize(1));
+        VSphereGuestInfoProperty guestInfoProperty = guestInfoProperties.get(0);
+        assertThat(guestInfoProperty, notNullValue());
+        assertThat(guestInfoProperty.getName(), is("JENKINS_URL"));
+        assertThat(guestInfoProperty.getValue(), is("${JENKINS_URL}"));
+        ComputerLauncher launcher = template.getLauncher();
+        assertThat(launcher, notNullValue());
+        assertThat(launcher, instanceOf(JNLPLauncher.class));
+        JNLPLauncher jnlpLauncher = (JNLPLauncher) launcher;
+        assertThat(jnlpLauncher.tunnel, is("jenkins:"));
+        RetentionStrategy<?> retentionStrategy = template.getRetentionStrategy();
+        assertThat(retentionStrategy, notNullValue());
+        assertThat(retentionStrategy, instanceOf(RunOnceCloudRetentionStrategy.class));
+        RunOnceCloudRetentionStrategy runOnce = (RunOnceCloudRetentionStrategy) retentionStrategy;
+        assertThat(runOnce.getIdleMinutes(), is(2));
+    }
+
+    @Test
+    public void should_support_configuration_export() throws Exception {
+        ConfiguratorRegistry registry = ConfiguratorRegistry.get();
+        ConfigurationContext context = new ConfigurationContext(registry);
+        final CNode cloud = getJenkinsRoot(context).get("clouds");
+
+        String exported = toYamlString(cloud);
+
+        String expected = toStringFromYamlFile(this, "expected_output.yml");
+
+        assertThat(exported, is(expected));
+    }
+}
--- a/src/test/resources/org/jenkinsci/plugins/vsphere/tools/configuration-as-code.yml
+++ b/src/test/resources/org/jenkinsci/plugins/vsphere/tools/configuration-as-code.yml
@@ -1,0 +1,40 @@
+jenkins:
+  systemMessage: "Hello World"
+  agentProtocols:
+    - "JNLP4-connect"
+  clouds:
+    - vSphere:
+        instanceCap: 100
+        maxOnlineSlaves: 0
+        templates:
+          - cloneNamePrefix: "windows-"
+            cluster: "Company"
+            datastore: "Company-FMD-01"
+            folder: "Company/Jenkins agents/Linked clones"
+            forceVMLaunch: true
+            guestInfoProperties:
+              # ^ escapes the secret
+              - name: "JENKINS_URL"
+                value: "^${JENKINS_URL}"
+            labelString: "windows vsphere"
+            launchDelay: 60
+            launcher:
+              jnlp:
+                tunnel: "jenkins:"
+            limitedRunCount: 1
+            linkedClone: true
+            masterImageName: "windows-server-2019"
+            mode: EXCLUSIVE
+            numberOfExecutors: 1
+            remoteFS: "C:/jenkins"
+            resourcePool: "Resources"
+            retentionStrategy:
+              runOnceCloud:
+                idleMinutes: 2
+            saveFailure: false
+            templateInstanceCap: 5
+            useSnapshot: true
+            waitForVMTools: true
+        vsConnectionConfig:
+          vsHost: "https://company-vsphere"
+        vsDescription: "Company vSphere"
--- a/src/test/resources/org/jenkinsci/plugins/vsphere/tools/expected_output.yml
+++ b/src/test/resources/org/jenkinsci/plugins/vsphere/tools/expected_output.yml
@@ -1,0 +1,34 @@
+- vSphere:
+    instanceCap: 100
+    maxOnlineSlaves: 0
+    templates:
+    - cloneNamePrefix: "windows-"
+      cluster: "Company"
+      datastore: "Company-FMD-01"
+      folder: "Company/Jenkins agents/Linked clones"
+      forceVMLaunch: true
+      guestInfoProperties:
+      - name: "JENKINS_URL"
+        value: "^${JENKINS_URL}"
+      labelString: "windows vsphere"
+      launchDelay: 60
+      launcher:
+        jnlp:
+          tunnel: "jenkins:"
+      limitedRunCount: 1
+      linkedClone: true
+      masterImageName: "windows-server-2019"
+      mode: EXCLUSIVE
+      numberOfExecutors: 1
+      remoteFS: "C:/jenkins"
+      resourcePool: "Resources"
+      retentionStrategy:
+        runOnceCloud:
+          idleMinutes: 2
+      saveFailure: false
+      templateInstanceCap: 5
+      useSnapshot: true
+      waitForVMTools: true
+    vsConnectionConfig:
+      vsHost: "https://company-vsphere"
+    vsDescription: "Company vSphere"


src/main/java/org/jenkinsci/plugins/vsphere/VSphereCloudRetentionStrategy.java
```diff
@@ -27,7 +26,7 @@ public DescriptorImpl getDescriptor() {
         return DESCRIPTOR;
     }
 
-    @Restricted(NoExternalUse.class)
+    @Extension
```
pjdarton: I am concerned that this might introduce a bug just like [this one in the docker plugin](https://github.com/jenkinsci/docker-plugin/issues/697), whose retention strategy [specifies it's an Extension](https://github.com/jenkinsci/docker-plugin/blob/docker-plugin-1.1.5/src/main/java/com/nirima/jenkins/plugins/docker/strategy/DockerOnceRetentionStrategy.java#L128) much as you have done.
The openstack-plugin's retention strategy has [this comment](https://github.com/jenkinsci/openstack-cloud-plugin/blob/openstack-cloud-2.41/src/main/java/jenkins/plugins/openstack/compute/JCloudsRetentionStrategy.java#L88) saying:
`// no @Extension since this retention strategy is used only for cloud nodes that we provision automatically.`
which makes me suspect that registering this as an Extension will cause this retention strategy to be offered to all manually-created slaves, not just vSphere ones.

Can you test this?

casz: Inside the descriptor, it should have a function to say which nodes this configuration is available for

pjdarton: I agree that, if an external plugin is to access this class, we'll have to drop the `@Restricted(NoExternalUse.class)` bit, but are you 100% sure it has to also be annotated with `@Extension`?
While I appreciate that it probably won't get automatically discovered without that annotation, it may be possible to specify it via a full package name (I don't know for sure - I've never used JCasc).

What we need to avoid is having Jenkins think it's a general-purpose retention strategy, so if it is marked as an extension, there will have to be some form of additional filtering to prevent it from being suggested as an option for non-vSphere slaves ... and I don't know enough about how this aspect of Jenkins works to suggest what we'd have to write in the plugin code to make that happen.
(all I do know is that, if we flag it as a generic retention strategy for general use, Jenkins will make it available for generic use, and bug reports will result)

casz: I'll look into it, since I need this for production since tomorrow 🤭

pjdarton: Well, FYI you can find a build of what you've created on the jenkinsci site, which is linked to within the "checks" shown below.
https://ci.jenkins.io/job/Plugins/job/vsphere-cloud-plugin/job/PR-100/ is the build for this PR - it has a build artifact of the hpi.

I suggest that you deploy that code for now and, when you have time, find out what's needed to enable your use-case without breaking it elsewhere - once we've got something that gives you what you need without unwanted side effects, I'd be happy to merge it in.

pjdarton: I think I've found out how it'd have to work - we'd need a DescriptorVisibilityFilter.
There's a list of implementations [here](https://jenkins.io/doc/developer/extensions/jenkins-core/#descriptorvisibilityfilter).
e.g. https://github.com/KostyaSha/yet-another-docker-plugin/blob/master/yet-another-docker-plugin/src/main/java/com/github/kostyasha/yad/DockerDescriptorVisibilityFilter.java

Our implementation would have to have code which said "if the descriptor is one of our retention strategies but the context is not docker then return false" (or something like that).

Disclaimer: I'm not an expert in this; I've just read a few bits of code, e.g. [Slave.java's retentionStrategyDescriptors method](https://github.com/jenkinsci/jenkins/blob/jenkins-2.155/core/src/main/java/hudson/model/Slave.java#L623) and [DescriptorVisibilityFilter](https://github.com/jenkinsci/jenkins/blob/jenkins-2.155/core/src/main/java/hudson/model/DescriptorVisibilityFilter.java).  I've never coded one of these before, or tested how this works...

casz: maybe @jenkinsci/code-reviewers have an idea for this 🤭

casz: DescriptorVisibilityFilterImpl seems to do the the trick 
image

pom.xml
```diff
@@ -42,6 +42,8 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <jenkins.version>2.150.1</jenkins.version>
```
casz: This is just for testing purpose, to get find those null pointers.

src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
```diff
@@ -175,8 +175,8 @@ public vSphereCloudSlaveTemplate(final String cloneNamePrefix,
         this.targetResourcePool = targetResourcePool;
         this.targetHost = targetHost;
         this.credentialsId = credentialsId;
-        this.nodeProperties = nodeProperties;
-        this.guestInfoProperties = guestInfoProperties;
+        this.nodeProperties = Util.fixNull(nodeProperties);
+        this.guestInfoProperties = Util.fixNull(guestInfoProperties);
```
pjdarton: While I'm kind-of ok with ensuring `nodeProperties` is an empty list instead of a null one, this isn't necessary for `guestInfoProperties` as the code which uses this copes with nulls.

...and if you want to ensure that either of these are non-null then we'll also need code in `readResolve()` to fix nulls there, as the constructor is only called when new instances are created; it isn't called when the instance gets de-serialized from the XML in config.xml
Personally, I quite like classes that get serialized as XML to have fields set to `null` when there's nothing in there (and then having the code handle nulls) as that way the serialized form of the class is smaller _and_ it makes the upgrade path easier, but I appreciate that it can make the code harder.
Either way, it's not a job for this PR; either another "generic code tidy-up" PR or not at all.

src/main/java/org/jenkinsci/plugins/vsphere/RunOnceCloudRetentionStrategy.java
```diff
@@ -176,4 +178,12 @@ public String getDisplayName() {
             return "vSphere Run-Once Retention Strategy";
         }
     }
+
+    @Extension
+    public static class DescriptorVisibilityFilterImpl extends DescriptorVisibilityFilter {
+        @Override
+        public boolean filter(@CheckForNull Object context, @NonNull Descriptor descriptor) {
+            return !(descriptor instanceof DescriptorImpl);
```
pjdarton: I don't see how this can be correct for all situations in which it'll be called.
My (possibly mistaken) understanding is that, when a user is manually creating a new (slave) Node, they can choose to create a vSphere node (with the ability to start/stop/revert-to-snapshot etc) and, in that `context` it'll be necessary to be able to select this strategy.
If `context` isn't a vSphere slave _then_ we don't allow this Descriptor through the filter, but I think we have to allow it through for vSphere slaves.

src/main/java/org/jenkinsci/plugins/vsphere/VSphereCloudRetentionStrategy.java
pjdarton: Hmm, looks like the Azure plugin has much the same problem that we're trying to fix here ;-)

src/main/java/org/jenkinsci/plugins/vsphere/builders/Deploy.java
```diff
@@ -318,7 +318,7 @@ private void add(String key, String val) {
         }
 
         @Override
-        public void buildEnvVars(AbstractBuild<?,?> build, EnvVars env) {
+        public void buildEnvironment(@Nonnull Run<?, ?> run, @Nonnull EnvVars env) {
```
pjdarton: This set of changes worries me.
There's a lot of tech-debt buried in these classes and, while it would be nice to resolve all of that, the testing effort to ensure that it still all works is non-trivial.
Best do that in a separate PR.

src/main/java/org/jenkinsci/plugins/vsphere/parameters/CloudSelectorParameter.java
```diff
@@ -1,21 +1,20 @@
...
+import java.util.List;
 import net.sf.json.JSONObject;
 import org.jenkinsci.plugins.vSphereCloud;
 import org.kohsuke.stapler.AncestorInPath;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.StaplerRequest;
 import org.kohsuke.stapler.export.Exported;
 
-import java.util.List;
+import static org.jenkinsci.plugins.vsphere.tools.PermissionUtils.throwUnlessUserHasPermissionToAccessJob;
```
pjdarton: Unnecessary code churn - this just makes it harder to merge the PR with other PRs.
In general, a PR should EITHER make functional changes (and keep code changes to the minimum required to do that) so that it's as easy as possible to do a manual merge with other changes OR make formatting changes (that don't affect functionality at all) so that folks can achieve the same effect on their own code knowing it changes nothing important - don't do both at once.
(personally, I prefer static imports at the top - I suspect your IDE is configured differently and did this on your behalf...)

pom.xml
```diff
@@ -3,11 +3,11 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.29</version>
+        <version>3.28</version>
     </parent>
 
     <artifactId>vsphere-cloud</artifactId>
-    <version>2.19-SNAPSHOT</version>
+    <version>3.0-SNAPSHOT</version>
```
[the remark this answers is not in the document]
casz: I know, I am just testing our infra and it was a quick way of signaling to my coworker that I was working on it.
I'll clean up the PR.

pom.xml
```diff
@@ -42,6 +42,8 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <jenkins.version>2.150.1</jenkins.version>
+        <java.level>8</java.level>
```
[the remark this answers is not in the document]
casz: Well to keep my head sane, I'll focus on java 8 and Jenkins 2.60.3 at the minimum but I properly would aim higher 😅

src/main/java/org/jenkinsci/plugins/vSphereCloudProvisionedSlave.java
```diff
@@ -134,16 +144,6 @@ public vSphereCloud getSpecificvSphereCloud(String vsDescription)
...
-        public List<Descriptor<ComputerLauncher>> getComputerLauncherDescriptors() {
-            List<Descriptor<ComputerLauncher>> result = new ArrayList<Descriptor<ComputerLauncher>>();
-            for (Descriptor<ComputerLauncher> launcher : Functions.getComputerLauncherDescriptors()) {
-                if (!vSphereCloudLauncher.class.isAssignableFrom(launcher.clazz)) {
-                    result.add(launcher);
-                }
-            }
-            return result;
-        }
-
```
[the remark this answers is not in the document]
casz: This is covered in the base class and descriptor filter covers the use case.

src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
casz: This is needed to be JCasc compliant, and I would rather NOT deal with potential null pointers. I feel it is safer be checking for empty lists.

src/main/java/org/jenkinsci/plugins/vSphereCloudLauncher.java
```diff
@@ -153,7 +153,7 @@ public void launch(SlaveComputer slaveComputer, TaskListener taskListener)
         {
             try {
 
-                if (slaveComputer.isTemporarilyOffline()) {
+                if (slaveComputer.isOffline()) {
```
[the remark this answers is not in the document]
casz: Doc says it all
https://javadoc.jenkins.io/hudson/model/Computer.html#isTemporarilyOffline--

src/main/java/org/jenkinsci/plugins/vSphereCloud.java
```diff
@@ -524,8 +524,7 @@ public Node call() throws Exception {
             };
             templateState.provisioningStarted(whatWeShouldSpinUp, nodeName);
             final Future<Node> provisionNodeTask = Computer.threadPoolForRemoting.submit(provisionNodeCallable);
-            final VSpherePlannedNode result = new VSpherePlannedNode(nodeName, provisionNodeTask, numberOfExecutors);
-            return result;
+            return new VSpherePlannedNode(nodeName, provisionNodeTask, numberOfExecutors);
```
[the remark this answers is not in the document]
casz: I applied some broad code cleanup from IDE, I'll revert 😅

src/main/java/org/jenkinsci/plugins/vSphereCloud.java
```diff
@@ -534,7 +533,7 @@ private static Node provisionNewNode(final CloudProvisioningRecord whatWeShouldS
             final vSphereCloudProvisionedSlave slave = template.provision(cloneName, StreamTaskListener.fromStdout());
             // ensure Jenkins knows about us before we forget what we're doing,
             // otherwise it'll just ask for more.
-            Jenkins.getInstance().addNode(slave);
+            Jenkins.get().addNode(slave);
```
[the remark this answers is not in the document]
casz: Everything is covered in javadoc, https://javadoc.jenkins-ci.org/jenkins/model/Jenkins.html#get-- in all of these cases it is .get when trying newer Jenkins.

src/main/java/org/jenkinsci/plugins/vSphereCloudSlave.java
```diff
@@ -365,6 +366,14 @@ public void preLaunch(Computer c, TaskListener taskListener) throws IOException,
...
     }
 
+    @Extension
+    public static class DescriptorVisibilityFilterImpl extends DescriptorVisibilityFilter {
+        @Override
+        public boolean filter(@CheckForNull Object context, @NonNull Descriptor descriptor) {
+            return !(descriptor instanceof DescriptorImpl);
+        }
+    }
+
```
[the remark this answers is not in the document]
casz: I'll look into, I was not aware you could manually provision a vSphere agent 😅

src/main/java/org/jenkinsci/plugins/vsphere/parameters/CloudSelectorParameter.java
casz: Yes, blame IDE 😅

src/main/java/org/jenkinsci/plugins/vsphere/builders/Deploy.java
casz: Again targeting newer Jenkins version this is the correct, the base class converts older AbstractBuild to Run for you behind the scene.

src/main/java/org/jenkinsci/plugins/vsphere/RunOnceCloudRetentionStrategy.java
casz: I'll look into the correct context.

src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
casz: When going via stapler and UI the node properties are definitely set to a empty list for you. But when JCasc configures the plugin it leaves them as null and then you should get null pointers somewhere down in the Slave/Computer object.

src/main/java/org/jenkinsci/plugins/vSphereCloud.java
```diff
@@ -362,11 +362,11 @@ private Integer calculateMaxAdditionalSlavesPermitted() {
                         + numberOfvSphereCloudSlaveExecutors + " executors), templates available are " + whatWeCouldUse);
                 while (excessWorkloadSoFar > 0) {
                     if (maxSlavesToProvisionBeforeCloudCapHit != null) {
-                        final int intValue = maxSlavesToProvisionBeforeCloudCapHit.intValue();
+                        final int intValue = maxSlavesToProvisionBeforeCloudCapHit;
```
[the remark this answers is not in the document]
casz: Code cleanup rules that my IDE applied 😅

src/main/java/org/jenkinsci/plugins/vSphereCloud.java
```diff
@@ -268,7 +268,7 @@ String getVsHost() {
     }
 
     public boolean getAllowUntrustedCertificate() {
-        return vsConnectionConfig != null ? vsConnectionConfig.getAllowUntrustedCertificate() : false;
+        return vsConnectionConfig != null && vsConnectionConfig.getAllowUntrustedCertificate();
```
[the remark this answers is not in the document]
casz: Code cleanup rules that my IDE applied 😅

casz:
@pjdarton I'll make it simple, this is very much WIP and I want something working today not in 6 months locally.

I will definitely cleanup anything as requested.

src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
pjdarton: Ah, OK, if this is required for JCasc then it can stay in this PR.
I don't think it's necessary to do this for `guestInfoProperties` as that code is fully null-proof, but if it's needed for `nodeProperties` then I guess that line is needed.
I would _prefer_ guestInfoProperties to be null if possible, as that means that when it's turned into XML there is no `<guestInfoProperties />` element, but if JCasc requires an empty list there as well then I guess it'll have to be `fixNull`ed too.

pom.xml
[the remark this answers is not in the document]
pjdarton: Well, if you want to do the work required to bring this plugin into the Jenkins-2 world that then that's OK with me, but I think it would be better done as a separate PR.
Personally, I would recommend bumping the Jenkins version to a fairly early Jenkins-2 (e.g. 2.60.3) because there might be folks "out there" using this plugin that are still using early versions.
e.g. where I work, we typically upgrade Jenkins once or twice a year (unless there's serious security concerns) - it already does what we need it to do, so there's little benefit to upgrading.

FYI If the files changed by doing that and by the JCasc changes don't overlap then it should be trivial to work on both PRs in parallel (which we'll only need to do until the Java8+Jenkins2.60.3 PR is merged in, which should only be a matter of days at most) and then the JCasc changes should auto-merge without additional human intervention.

src/main/java/org/jenkinsci/plugins/vSphereCloud.java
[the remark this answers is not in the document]
pjdarton: I think you need to "have words" with your IDE and tell it not to go fiddling unnecessarily :-)
I would also suggest that you get in the habit of checking all the code changes you're about to make prior to committing/pushing them - IDEs don't always get things right.

src/main/java/org/jenkinsci/plugins/vSphereCloud.java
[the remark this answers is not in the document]
pjdarton: OK, I've read the javadocs and looked at the code...
It seems to me that the direct like-for-like replacement for `Jenkins.getInstance()` is `Jenkins.getInstanceOrNull()` and the direct like-for-like replacement for `Jenkins.getActiveInstance()` is `Jenkins.get()`.
However, in any situation in the code where it's currently calling `Jenkins.getInstance()` and where it is _not_ checking for a `null` result then that code is wrong and should really have been calling `Jenkins.getActiveInstance()` instead because that method handles the null situation much better than a generic NPE.
i.e. there's some tech-debt here.

The problem is that `Jenkins.get()` says it was introduced in Jenkins 2.98, and it wasn't back-ported to LTS, so the first LTS release of Jenkins that contained that was version 2.107.1 which was released on 14th March 2018 and that's not old enough for my liking - I think we should give folks at least a year to upgrade.

So, I think that we should hold off switching to using `Jenkins.get()` for now as that's a bit _too_ new ... but perhaps after we've updated to require Jenkins 2.60.3 then we could replace all calls to `Jenkins.getInstance()` with either `Jenkins.getInstanceOrNull()` or `Jenkins.getActiveInstance()` so that the code is 100% clear what it needs in each situation (and would make a later update to use `Jenkins.get()` much more straightforward).

src/main/java/org/jenkinsci/plugins/vSphereCloudLauncher.java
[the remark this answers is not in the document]
pjdarton: According to the docs, these two are _not_ 100% equivalent: isTemporarilyOffline() says "In contrast, isOffline() represents the actual online/offline state. For example, this method may return false while isOffline() returns true if the agent failed to launch."

I'm not convinced that this code change is correct - I suspect that the code here really is more interested in whether or not there is a user-specified reason for the node being kept offline rather than it just being "accidentally" offline.
As the docs say "This method is marked as deprecated to warn people when they accidentally call this method." - the deprecation warning doesn't mean it's always wrong to call it, it just means that one has to be careful in choosing which of the two methods one means.

Personally, I would advise you to avoid changing this class unless you absolutely have to.  Unless you are 100% sure that you're fixing a bug by making this change, it's better to leave it unchanged as _any_ changes here can just as easily introduce a bug as fix one.

src/main/java/org/jenkinsci/plugins/vSphereCloudSlave.java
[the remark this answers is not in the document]
pjdarton: It's not a use-case I make any use of myself either, but a lot of people do.
e.g. having a set of (static) VMs with a snapshot on them where Jenkins reverts the VM to that snapshot after every build.

FYI That's where most of the different `vSphereCloudLauncher.MACHINE_ACTION` options come in useful (as they're not very relevant for an automatically provisioned machine which is disposed of after the build completes).

src/main/java/org/jenkinsci/plugins/vsphere/builders/Deploy.java
pjdarton: OK.  There's a fair amount of code duplication tech-debt in this area and it'd be nice to get that cleaned up ... but lets do that in a separate PR after bumping the Java version and minimum Jenkins version.

pjdarton:
OK, if you need "something working today" then I think we may need to handle this a little differently.
I would suggest that you continue to use this PR as your own work area, but on the understanding that this PR isn't going to get merged as-is.
Once the immediate urgency is over, we can work on doing this as a series of small limited-scope PRs _without_ the same time pressures:
1) Update to Java8 and Jenkins 2.60.3
2) Code cleanup: Stop using `Jenkins.getInstance()`, tidy up the builders code to use `Run` instead of `AbstractBuild`, check any other deprecation warnings, take advantage of Java8 syntax etc.
3) Make the plugin JCasc-compliant by making classes `@Extension` and adding filters.

FYI I went through a similar process when working on jenkins plugins - what starts out as a "simple change" turns into a serious amount of tidying up and refactoring that's all required before one can apply the original "simple change".  I found it was easier to pull things into separate PRs and then rebase the "everything combined" PR (and thus see progressively fewer changes as the master branch got closer to my "combined" branch), and this allowed each set of changes to be reviewed and tested in isolation.

casz:
I need to get back to this 😓 

vSphere plugin has been working great for us with these additions.

casz:
Taken the first few steps towards this: #110

casz:
I will rebase onto #110 and unit testing for JCasC :)

pom.xml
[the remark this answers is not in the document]
casz: Being done in a separate PR

src/main/java/org/jenkinsci/plugins/vsphere/parameters/CloudSelectorParameter.java
```diff
@@ -35,8 +35,10 @@ public StringParameterValue getDefaultParameterValue() {
 
     private StringParameterValue checkValue(StringParameterValue value) {
         List<String> cloudNames = vSphereCloud.findAllVsphereCloudNames();
-        if (!cloudNames.contains(value.value))
-            throw new IllegalArgumentException("No vsphere cloud with name: " + value.value);
+        String check = String.valueOf(value.getValue());
```
casz: reason for this change is that `value.value` is illegal to access in recent version of Jenkins.

casz:
I believe #110 is ready for merge and #100 I will add JCasC test shortly.

I tested my JCasC test (unit test) against the `pom` by cherry picking branch (locally) and my `master` branch I can confirm that `VSphereCloudRetentionStrategy` needs to be marked as an extension point:

```
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.185 s <<< FAILURE! - in org.jenkinsci.plugins.vsphere.tools.ConfigurationAsCodeTest
[ERROR] org.jenkinsci.plugins.vsphere.tools.ConfigurationAsCodeTest  Time elapsed: 1.182 s  <<< FAILURE!
java.lang.AssertionError: Unexpected exception
        at io.jenkins.plugins.casc.misc.JenkinsConfiguredWithCodeRule.before(JenkinsConfiguredWithCodeRule.java:37)
        at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:593)
        at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:298)
        at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:292)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.IllegalArgumentException: No hudson.slaves.RetentionStrategy implementation found for runOnceCloud         at io.jenkins.plugins.casc.impl.configurators.HeteroDescribableConfigurator.lambda$lookupDescriptor$11(HeteroDescribableConfigurator.java:212)
        at io.vavr.control.Option.orElse(Option.java:321)
        at io.jenkins.plugins.casc.impl.configurators.HeteroDescribableConfigurator.lookupDescriptor(HeteroDescribableConfigurator.java:211)
        at io.jenkins.plugins.casc.impl.configurators.HeteroDescribableConfigurator.lambda$configure$3(HeteroDescribableConfigurator.java:84)
        at io.vavr.Tuple2.apply(Tuple2.java:238)
        at io.jenkins.plugins.casc.impl.configurators.HeteroDescribableConfigurator.configure(HeteroDescribableConfigurator.java:83)
        at io.jenkins.plugins.casc.impl.configurators.HeteroDescribableConfigurator.configure(HeteroDescribableConfigurator.java:55)
        at io.jenkins.plugins.casc.impl.configurators.DataBoundConfigurator.tryConstructor(DataBoundConfigurator.java:160)
        at io.jenkins.plugins.casc.impl.configurators.DataBoundConfigurator.instance(DataBoundConfigurator.java:77)             at io.jenkins.plugins.casc.BaseConfigurator.configure(BaseConfigurator.java:267)
        at io.jenkins.plugins.casc.impl.configurators.DataBoundConfigurator.configure(DataBoundConfigurator.java:83)
        at io.jenkins.plugins.casc.impl.configurators.DataBoundConfigurator.tryConstructor(DataBoundConfigurator.java:152)
        at io.jenkins.plugins.casc.impl.configurators.DataBoundConfigurator.instance(DataBoundConfigurator.java:77)
        at io.jenkins.plugins.casc.BaseConfigurator.configure(BaseConfigurator.java:267)
        at io.jenkins.plugins.casc.impl.configurators.DataBoundConfigurator.configure(DataBoundConfigurator.java:83)
        at io.jenkins.plugins.casc.impl.configurators.HeteroDescribableConfigurator.lambda$doConfigure$16668e2$1(HeteroDescribableConfigurator.java:278)
        at io.vavr.CheckedFunction0.lambda$unchecked$52349c75$1(CheckedFunction0.java:247)
        at io.jenkins.plugins.casc.impl.configurators.HeteroDescribableConfigurator.doConfigure(HeteroDescribableConfigurator.java:278)
        at io.jenkins.plugins.casc.impl.configurators.HeteroDescribableConfigurator.lambda$null$2(HeteroDescribableConfigurator.java:86)
        at io.vavr.control.Option.map(Option.java:392)
        at io.jenkins.plugins.casc.impl.configurators.HeteroDescribableConfigurator.lambda$configure$3(HeteroDescribableConfigurator.java:86)
        at io.vavr.Tuple2.apply(Tuple2.java:238)
        at io.jenkins.plugins.casc.impl.configurators.HeteroDescribableConfigurator.configure(HeteroDescribableConfigurator.java:83)
        at io.jenkins.plugins.casc.impl.configurators.HeteroDescribableConfigurator.check(HeteroDescribableConfigurator.java:92)
        at io.jenkins.plugins.casc.impl.configurators.HeteroDescribableConfigurator.check(HeteroDescribableConfigurator.java:55)
        at io.jenkins.plugins.casc.BaseConfigurator.configure(BaseConfigurator.java:344)
        at io.jenkins.plugins.casc.BaseConfigurator.check(BaseConfigurator.java:287)
        at io.jenkins.plugins.casc.ConfigurationAsCode.lambda$checkWith$7(ConfigurationAsCode.java:716)
        at io.jenkins.plugins.casc.ConfigurationAsCode.invokeWith(ConfigurationAsCode.java:666)
        at io.jenkins.plugins.casc.ConfigurationAsCode.checkWith(ConfigurationAsCode.java:716)
        at io.jenkins.plugins.casc.ConfigurationAsCode.configureWith(ConfigurationAsCode.java:701)
        at io.jenkins.plugins.casc.ConfigurationAsCode.configureWith(ConfigurationAsCode.java:592)
        at io.jenkins.plugins.casc.ConfigurationAsCode.configure(ConfigurationAsCode.java:562)
        at io.jenkins.plugins.casc.misc.JenkinsConfiguredWithCodeRule.before(JenkinsConfiguredWithCodeRule.java:34)             ... 5 more
```

casz:
This should definitely be a squash and merge 😅

src/main/java/org/jenkinsci/plugins/vSphereCloud.java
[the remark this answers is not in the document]
casz: reverted

src/main/java/org/jenkinsci/plugins/vSphereCloud.java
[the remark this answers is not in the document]
casz: reverted

src/main/java/org/jenkinsci/plugins/vSphereCloud.java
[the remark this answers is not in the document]
casz: reverted

src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
casz: Still needed otherwise user will experience null pointer exception when applying JCasC configuration

src/main/java/org/jenkinsci/plugins/vsphere/parameters/CloudSelectorParameter.java
casz: reverted

src/main/java/org/jenkinsci/plugins/vsphere/builders/Deploy.java
casz: Reverted

casz:
There @pjdarton now #100 is also ready for merging.

Feel free to merge #110 and then I will rebase #100 😄

src/main/java/org/jenkinsci/plugins/vsphere/VSphereCloudRetentionStrategy.java
casz: Definitely still need the extension, no need to remove the restricted annotation.

casz:
This is the better change set to read: https://github.com/casz/vsphere-cloud-plugin/compare/casz:pom...master

casz:
using the incremental, that is produced from here: I am seeing `No hudson.slaves.RetentionStrategy implementation found for vSphereCloud` which is odd.

As the JCasC unit test is testing that very same thing 😕

pjdarton:
This is just a guess ... but in my experience, when a local build works and a server build gives a "X not found" error, the local build wasn't "clean" and contained leftovers from a previous build that allowed it to work.

casz:
mvn clean deploy 🤔 should clean the target.
Also Jenkins CI passes the unit test.

src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
```diff
@@ -507,7 +507,6 @@ private ComputerLauncher determineLauncher(final VSphere vSphere, final String c
                     sshLauncher.getPrefixStartSlaveCmd(), sshLauncher.getSuffixStartSlaveCmd(),
                     sshLauncher.getLaunchTimeoutSeconds(), sshLauncher.getMaxNumRetries(),
                     sshLauncher.getRetryWaitTime(), sshLauncher.getSshHostKeyVerificationStrategy());
-            return launcherWithIPAddress;
```
pjdarton: I must've missed this first time around, but this doesn't look right to me - that'll mean nobody can use a SSHLauncher to connect the slave to Jenkins.

casz: not sure what happened here, must be one of my many merge conflicts that has gone bad.

casz: ah, revert cleanup failed to cleanup this merge conflict 😆

casz: Cleaned up the git history

casz: Sorry for the confusion, I confused myself 😭

casz:
Latest incremental is A OKAY 🙌 Perhaps the plugin was not successfully activated in our staging environment.

So from my point it is 🚢 📦 🚀

pjdarton:
Hmm... I would merge this, but github ignored my previous "Confirm squash and merge" button-press, gave me an error 500 on another operation and now it's saying "Checking for ability to merge automatically…" which doesn't smell right to me.
At work, there were reports of our GitHubEnterprise service (which is run by gitHub) having issues so maybe there's a common cause...

casz:
let me merge with upstream master

casz:
@pjanouse https://www.githubstatus.com/ GitHub seems to have some degraded statuses

pjdarton:
Either you did something, or github changed its mind...

pjdarton:
FYI #112 now has merge issues.
If you can resolve those (I'd guess it'll be fairly simple) and add a PR description (and you agree with my reasoning in my comment) then I'd be happy to merge that too.

casz:
You can probably remove wip label 😂🔖

PR #103, opened 2019-03-09
Add additional parameters to reconfiguring VMs

nakkapeddi:
Howdy folks,

According to the vSphere WSSDK docs for [vim.vm.ConfigSpec](https://pubs.vmware.com/vi3/sdk/ReferenceGuide/vim.vm.ConfigSpec.html), you can setup reservations and limits for both CPU and memory.

In the course of my work, I've had to specify a CPU reservation limit for my linked clone build agent VMs, and ended up forking the plugin to achieve this. Are there any plans to support the additional config changes to the spec? I rewrote the existing default constructor, but ideally I'd create a new one to support the additional config parameters. Thoughts?

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/vsphere/builders/ReconfigureCpu.java
+++ b/src/main/java/org/jenkinsci/plugins/vsphere/builders/ReconfigureCpu.java
@@ -24,6 +24,7 @@ import org.jenkinsci.plugins.vsphere.tools.VSphereException;
 import org.jenkinsci.plugins.vsphere.tools.VSphereLogger;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.QueryParameter;
+import com.vmware.vim25.ResourceAllocationInfo;
 
 import javax.annotation.Nonnull;
 import javax.servlet.ServletException;
@@ -34,11 +35,16 @@ public class ReconfigureCpu extends ReconfigureStep {
 
     private final String cpuCores;
     private final String coresPerSocket;
+    private final String cpuLimitMHz;
+    private final ResourceAllocationInfo cpuReservation;
 
 	@DataBoundConstructor
-	public ReconfigureCpu(String cpuCores, String coresPerSocket) throws VSphereException {
+	public ReconfigureCpu(String cpuCores, String coresPerSocket, String cpuLimitMHz) throws VSphereException {
 		this.cpuCores = cpuCores;
         this.coresPerSocket = coresPerSocket;
+        this.cpuLimitMHz = cpuLimitMHz;
+        this.cpuReservation = new ResourceAllocationInfo();
+        this.cpuReservation.setReservation((long)Integer.valueOf(this.cpuLimitMHz));
 	}
 
 	public String getCpuCores() {
@@ -75,6 +81,8 @@ public class ReconfigureCpu extends ReconfigureStep {
         PrintStream jLogger = listener.getLogger();
         String expandedCPUCores = cpuCores;
         String expandedCoresPerSocket = coresPerSocket;
+        ResourceAllocationInfo resAllInfo = cpuReservation;
+
         EnvVars env;
         try {
             env = run.getEnvironment(listener);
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/vsphere/builders/ReconfigureCpu.java
```diff
@@ -34,11 +35,16 @@
 
     private final String cpuCores;
     private final String coresPerSocket;
+    private final String cpuLimitMHz;
+    private final ResourceAllocationInfo cpuReservation;
```
pjdarton: You've introduced data duplication here.
EITHER store the `cpuLimitMHz` only OR store the `cpuReservation` only ... but _do not_ store a `cpuReservation` that (always) contains a limit of `cpuLimitMHz`.

Personally, I'd recommend having only a field of `cpuLimitMHz` and pushing the code that creates a `ResourceAllocationInfo` down to `reconfigureCPU` where it's used.

src/main/java/org/jenkinsci/plugins/vsphere/builders/ReconfigureCpu.java
```diff
@@ -34,11 +35,16 @@
...
 
 	@DataBoundConstructor
-	public ReconfigureCpu(String cpuCores, String coresPerSocket) throws VSphereException {
+	public ReconfigureCpu(String cpuCores, String coresPerSocket, String cpuLimitMHz) throws VSphereException {
```
pjdarton: Don't change the constructor - that can break things for existing users, especially folks who're using pipeline builds.
Create a `@DataBoundSetter` method `setCpuLimitMHz()` instead.
(Yes, this will mean that `cpuLimitMHz` can't be final - just put a comment in saying `/* almost final */`)

FYI while this plugin pre-dates such a practice being widely known, these days it's considered Jenkins "best practice" to have a constructor that only take mandatory arguments (and to have as few of those as possible) with everything else being set by data bound set methods.  This makes it much easier for code that doesn't know or care about newly introduced optional fields to continue to work.

src/main/java/org/jenkinsci/plugins/vsphere/builders/ReconfigureCpu.java
```diff
@@ -75,6 +81,8 @@ public boolean reconfigureCPU (final Run<?, ?> run, final Launcher launcher, fin
         PrintStream jLogger = listener.getLogger();
         String expandedCPUCores = cpuCores;
         String expandedCoresPerSocket = coresPerSocket;
+        ResourceAllocationInfo resAllInfo = cpuReservation;
```
pjdarton: Replace with something like
```
ResourceAllocationInfo resAllInfo = new ResourceAllocationInfo();
cpuReservation.setReservation((long)Integer.valueOf(getCpuLimitMHz()));
```
...except that you'll also need to make sure that, if `cpuLimitMHz` is null or empty (which it can be for folks who aren't using this new capability), you don't call `spec.setCpuAllocation(resAllInfo)` at all.

src/main/java/org/jenkinsci/plugins/vsphere/builders/ReconfigureCpu.java
```diff
@@ -91,6 +99,7 @@ public boolean reconfigureCPU (final Run<?, ?> run, final Launcher launcher, fin
         VSphereLogger.vsLogger(jLogger, "Preparing reconfigure: CPU");
         spec.setNumCPUs(Integer.valueOf(expandedCPUCores));
         spec.setNumCoresPerSocket(Integer.valueOf(expandedCoresPerSocket));
+        spec.setCpuAllocation(resAllInfo);
```
pjdarton: If the user hasn't set `cpuLimitMHz` then `spec.setCpuAllocation(...)` should not be called.
i.e. preserve existing default behavior.

pjdarton:
@nakkapeddi I'd like to do a new release before the middle of next month (there are other changes awaiting release) so time is running out to resolve the changes requested...

If anything is unclear, just ask; I'm happy to advise.

pjdarton:
This PR has been inactive for some time now...
@nakkapeddi are you still interested in completing this and getting it merged in?

I'm not keen on having PRs open for ages - I'd prefer that we either work on them until they're complete and get them merged in, or we close them if that's not going to happen.

PR #104, opened 2019-06-26
[JENKINS-57967] Pass SshHostKeyVerificationStrategy to SSHLauncher

medianick:
Previous version of the constructor was removed in SSH Slaves 1.30.0. Fixes problem detailed in https://issues.jenkins-ci.org/browse/JENKINS-57967.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,7 +80,7 @@
          <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>ssh-slaves</artifactId>
-            <version>1.11</version>
+            <version>1.15</version>
          </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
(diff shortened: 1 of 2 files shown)


pjdarton:
Removed in 1.30.0?  Ah, I guess I best not update that plugin on my vSphere-using Jenkins servers then...

Once you're happy that this PR is complete, let me know; we can then both test the results in our respective "real world" environments and, if that all looks good, I'll kick out a new release...

medianick:
I believe this PR is complete; it builds successfully here, and I've also installed it on my test Jenkins system, upgraded SSH Slaves to 1.30.0 there, and was able to spawn and build on all the different ssh-connected vSphere  templates I have set up. It's my first time contributing to this plugin, so if I've missed any typical steps, please let me know.

pjdarton:
1st-time or not, it looks like you've done everything right.
 - Kept code churn to a minimum too, which makes my life easier ;-)
 - Tested it in your own environment to make sure it works :-)
 - Flagged it as "draft" until you were ready.
The only suggestion I have about any future contributions is that the PR description at the top is very terse - I had to see what the code changes were before I really understood what this was about.  That wasn't insurmountable this time (as the code changes are minimal) but if you ever tackle anything requiring more than a 3-line change then a full description of the problem the PR is solving would be appreciated.

Out of interest, what was the symptom(s) of upgrading the SSH plugin to 1.30.0 _without_ this fix?
I guess there will be a nasty exception and a failure to provision a VM ... but it'd be nice to record what that exception is so that anyone searching for that exception will then find this PR and know what they need to do...

medianick:
@pjdarton -- thank you. I'd added the link to the corresponding Jira ticket in the body of my first commit message -- meaning for that to become part of the PR description -- but then did a catch-up with origin first, which resulted in needing to fill in the PR description manually, so I lost that. I've added it back. You can see the stack trace, etc. on that ticket.

pjdarton:
Plugin version 2.20 has been built & uploaded; it should be on the update servers soon.

PR #107, opened 2019-09-03
Adding LICENSE file to project

kshithijiyer:


pjdarton:
I've been looking into this some more and I don't believe that this file is entirely necessary.
The `pom.xml` file specifies the Apache 2.0 license applies to the plugin code overall (and this is what Jenkins reports in the plugins pages).
There's no need to copy/paste the full license text here as that just adds data-duplication.

If you want to make it more obvious for people what the license is, I'd suggest adding a `LICENSE` file which simply instructs readers to look in the pom.xml file, as that's where the definitive statement is.

pjdarton:
This has been an interesting PR that provoked some investigations into unfamiliar (for me at least) territory.  However, we've come out of it stronger...

Because of the questions this raised, I pinged my friendly lawyer contact and they "suggested" that we really ought to make it very clear (a) what the license was and (b) that any contributions to this code had to also be made under the same license.
So, I did.
As this PR hadn't been updated to include the truncated LICENSE file I requested, I couldn't use this PR for the basis of all that work, so I created #109 instead.

I believe that #109 addresses all the concerns that prompted this one, and goes beyond that to make it abundantly clear precisely what software license this code may be used under ... as well as providing some (long overdue!) suggestions about how best to go about contributing.

So, thanks for the PR - it certainly pushed things forward - but it's now been overtaken by the changes it prompted and is now obsolete.

PR #110, opened 2019-10-24
prepare for JCasC compatibility

casz:
- update parent pom to latest v3.51
- set java level to 8.
- target Jenkins 2.60.3 in preparation for adding JCasC compatibility
- fix upper bound dependency between credential plugins and ssh slaves plugin
- json-lib is coming from Jenkins core.
- remove custom build steps as they are no longer necessary due to pom update
  and spotbugs.skip is available
- fix credential deprecation warnings
- fix incompatible type after jenknis 2.60.3 and avoid raw types
- fix tabs/indents in pom 😭

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,8 +65,8 @@
         <spotbugs.skip>true</spotbugs.skip>
     </properties>
 
-        <dependencies>
-            <dependency>
+    <dependencies>
+        <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>credentials</artifactId>
             <version>2.1.2</version>
@@ -83,16 +83,16 @@
             <artifactId>yavijava</artifactId>
             <version>6.0.05</version>
         </dependency>
-         <dependency>
-             <groupId>org.jenkins-ci.plugins</groupId>
-             <artifactId>node-iterator-api</artifactId>
-             <version>1.5</version>
-         </dependency>
-         <dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>node-iterator-api</artifactId>
+            <version>1.5</version>
+        </dependency>
+        <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>ssh-slaves</artifactId>
             <version>1.15</version>
-         </dependency>
+        </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-basic-steps</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
     </parent>
 
     <artifactId>vsphere-cloud</artifactId>
-    <version>2.22-SNAPSHOT</version>
+    <version>${revision}${changelist}</version>
     <packaging>hpi</packaging>
 
     <name>vSphere Plugin</name>
@@ -54,10 +54,12 @@
         <connection>scm:git:git://github.com/jenkinsci/vsphere-cloud-plugin.git</connection>
         <developerConnection>scm:git:ssh://git@github.com/jenkinsci/vsphere-cloud-plugin.git</developerConnection>
         <url>https://github.com/jenkinsci/vsphere-cloud-plugin</url>
-        <tag>HEAD</tag>
+        <tag>${scmTag}</tag>
     </scm>
 
     <properties>
+        <revision>2.22</revision>
+        <changelist>-SNAPSHOT</changelist>
         <revision>2.21</revision>
         <changelist>-SNAPSHOT</changelist>
         <java.level>8</java.level>
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.29</version>
+        <version>3.51</version>
     </parent>
 
     <artifactId>vsphere-cloud</artifactId>
@@ -60,20 +60,16 @@
     <properties>
         <revision>2.21</revision>
         <changelist>-SNAPSHOT</changelist>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <java.level>8</java.level>
+        <jenkins.version>2.60.3</jenkins.version>
+        <spotbugs.skip>true</spotbugs.skip>
     </properties>
 
-    <dependencies>
-        <dependency>
-            <groupId>org.kohsuke.stapler</groupId>
-            <artifactId>json-lib</artifactId>
-            <version>2.1-rev7</version>
-            <type>jar</type>
-        </dependency>
-        <dependency>
+        <dependencies>
+            <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>credentials</artifactId>
-            <version>1.28</version>
+            <version>2.1.2</version>
             <type>jar</type>
         </dependency>
         <dependency>
@@ -116,23 +112,4 @@
             <url>https://repo.jenkins-ci.org/public/</url>
         </pluginRepository>
     </pluginRepositories>
-
-    <build>
-        <plugins>
-            <plugin>
-                <groupId>org.jenkins-ci.tools</groupId>
-                <artifactId>maven-hpi-plugin</artifactId>
-                <configuration>
-                    <disabledTestInjection>false</disabledTestInjection>
-                </configuration>
-            </plugin>
-            <plugin>
-                <groupId>org.codehaus.mojo</groupId>
-                <artifactId>findbugs-maven-plugin</artifactId>
-                <configuration>
-                    <skip>true</skip>
-                </configuration>
-            </plugin>
-        </plugins>
-    </build>
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
     </parent>
 
     <artifactId>vsphere-cloud</artifactId>
-    <version>${revision}${changelist}</version>
+    <version>2.22-SNAPSHOT</version>
     <packaging>hpi</packaging>
 
     <name>vSphere Plugin</name>
@@ -54,12 +54,10 @@
         <connection>scm:git:git://github.com/jenkinsci/vsphere-cloud-plugin.git</connection>
         <developerConnection>scm:git:ssh://git@github.com/jenkinsci/vsphere-cloud-plugin.git</developerConnection>
         <url>https://github.com/jenkinsci/vsphere-cloud-plugin</url>
-        <tag>${scmTag}</tag>
+        <tag>HEAD</tag>
     </scm>
 
     <properties>
-        <revision>2.22</revision>
-        <changelist>-SNAPSHOT</changelist>
         <revision>2.21</revision>
         <changelist>-SNAPSHOT</changelist>
         <java.level>8</java.level>
--- a/src/main/java/org/jenkinsci/plugins/vSphereCloudSlave.java
+++ b/src/main/java/org/jenkinsci/plugins/vSphereCloudSlave.java
@@ -371,7 +371,7 @@ public class vSphereCloudSlave extends AbstractCloudSlave {
     public static class DescriptorVisibilityFilterImpl extends DescriptorVisibilityFilter {
         @Override
         public boolean filter(@CheckForNull Object context, @NonNull Descriptor descriptor) {
-            return !(descriptor instanceof vSphereCloudProvisionedSlave.DescriptorImpl);
+            return !(descriptor instanceof DescriptorImpl);
         }
     }
 
--- a/src/main/java/org/jenkinsci/plugins/vSphereCloudSlave.java
+++ b/src/main/java/org/jenkinsci/plugins/vSphereCloudSlave.java
@@ -2,10 +2,12 @@ package org.jenkinsci.plugins;
 
 import static org.jenkinsci.plugins.vsphere.tools.PermissionUtils.throwUnlessUserHasPermissionToConfigureSlave;
 
+import edu.umd.cs.findbugs.annotations.CheckForNull;
+import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.AbortException;
 import hudson.Extension;
-import hudson.Functions;
 import hudson.Util;
+import hudson.model.DescriptorVisibilityFilter;
 import hudson.model.Queue.BuildableItem;
 import hudson.model.Result;
 import hudson.model.TaskListener;
@@ -366,6 +368,14 @@ public class vSphereCloudSlave extends AbstractCloudSlave {
     }
 
     @Extension
+    public static class DescriptorVisibilityFilterImpl extends DescriptorVisibilityFilter {
+        @Override
+        public boolean filter(@CheckForNull Object context, @NonNull Descriptor descriptor) {
+            return !(descriptor instanceof vSphereCloudProvisionedSlave.DescriptorImpl);
+        }
+    }
+
+    @Extension
     public static final class DescriptorImpl extends SlaveDescriptor {
 
         public DescriptorImpl() {
@@ -400,16 +410,6 @@ public class vSphereCloudSlave extends AbstractCloudSlave {
                 }
             }
             throw new Exception("The vSphere Cloud doesn't exist");
-        }
-
-        public List<Descriptor<ComputerLauncher>> getComputerLauncherDescriptors() {
-            List<Descriptor<ComputerLauncher>> result = new ArrayList<Descriptor<ComputerLauncher>>();
-            for (Descriptor<ComputerLauncher> launcher : Functions.getComputerLauncherDescriptors()) {
-                if (!vSphereCloudLauncher.class.isAssignableFrom(launcher.clazz)) {
-                    result.add(launcher);
-                }
-            }
-            return result;
         }
 
         public List<String> getIdleOptions() {
--- a/src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
+++ b/src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
@@ -18,6 +18,7 @@ package org.jenkinsci.plugins;
 
 import static org.jenkinsci.plugins.vsphere.tools.PermissionUtils.throwUnlessUserHasPermissionToConfigureCloud;
 
+import hudson.DescriptorExtensionList;
 import hudson.EnvVars;
 import hudson.Extension;
 import hudson.Util;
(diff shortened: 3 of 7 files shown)


casz:
Jenkins says NO, need to add the DescriptorVisibilityFilterImpl from #100 😄 

https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fvsphere-cloud-plugin/detail/PR-110/1/pipeline/#step-31-log-44

casz:
Here is the code related to visibility 

```java
        /**
         * Returns the list of {@link ComputerLauncher} descriptors appropriate to the supplied {@link Slave}.
         *
         * @param it the {@link Slave} or {@code null} to assume the slave is of type {@link #clazz}.
         * @return the filtered list
         * @since 2.12
         */
        @Nonnull
        @Restricted(NoExternalUse.class) // intended for use by Jelly EL only (plus hack in DelegatingComputerLauncher)
        public final List<Descriptor<ComputerLauncher>> computerLauncherDescriptors(@CheckForNull Slave it) {
            DescriptorExtensionList<ComputerLauncher, Descriptor<ComputerLauncher>> all =
                    Jenkins.getInstance().<ComputerLauncher, Descriptor<ComputerLauncher>>getDescriptorList(
                            ComputerLauncher.class);
            return it == null ? DescriptorVisibilityFilter.applyType(clazz, all)
                    : DescriptorVisibilityFilter.apply(it, all);
        }
```

https://github.com/jenkinsci/jenkins/blob/81e52b83d5caefd9a9637b47787baf9cfddb0421/core/src/main/java/hudson/model/Slave.java#L544-L559

Does the same as what 

https://github.com/jenkinsci/vsphere-cloud-plugin/blob/8e39d601d5c96be819fe22dd5ecd68bd481cd57e/src/main/java/org/jenkinsci/plugins/vSphereCloudProvisionedSlave.java#L137-L145

casz:
Okay green build 👍 

Should I apply the DescriptorVisibilityFilter to the other methods here or do you want a separate PR for that?

src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
```diff
@@ -645,15 +647,10 @@ public FormValidation doTestCloneParameters(@AncestorInPath AbstractFolder<?> co
...
-            for (NodePropertyDescriptor npd : list) {
-                if (npd.isApplicable(vSphereCloudSlave.class)) {
-                    result.add(npd);
-                }
-            }
-            return result;
+            DescriptorExtensionList<NodeProperty<?>, NodePropertyDescriptor> nodePropertyDescriptors = NodeProperty.all();
+            return nodePropertyDescriptors.stream()
+                .filter(npd -> npd.isApplicable(vSphereCloudSlave.class))
+                .collect(Collectors.toList());
```
pjdarton: I'm not hugely keen of the aesthetics on the syntax here but I presume that this is Java-8 syntax for applying the same filter?
(I'm not asking for changes, just reassurances that this is "how it's normally done these days" as I've not seen that code pattern before)

src/main/java/org/jenkinsci/plugins/vsphere/VSphereConnectionConfig.java
```diff
@@ -140,14 +141,9 @@ public FormValidation doCheckAllowUntrustedCertificate(@QueryParameter boolean v
...
-            if (instance != null) {
-                return new StandardListBoxModel().withEmptySelection().withMatching(
-                    CREDENTIALS_MATCHER, CredentialsProvider.lookupCredentials(StandardCredentials.class,
-                    instance, ACL.SYSTEM, getDomainRequirement(vsHost))
-                );
-            }
-            return new StandardListBoxModel();
+            return new StandardListBoxModel().includeEmptyValue()
+                .includeMatching(Jenkins.getInstance(), StandardCredentials.class,
+                    Collections.singletonList(getDomainRequirement(vsHost)), CREDENTIALS_MATCHER);
```
pjdarton: :+1:

pjdarton:
Thanks for getting back on this :-)

FYI my "at work" workload is currently such that I don't have the time to deep delve into this and learn enough to offer informed-and-instructive opinion so I guess you're going to be the subject-matter-expert on this without much help from me (although, if you can point me at specific documentation to improve my understanding, that'd be appreciated).
...also FYI, since you originally submitted #100, I added some [contribution guidelines](https://github.com/jenkinsci/vsphere-cloud-plugin/blob/master/CONTRIBUTING.md) which I'd suggest you check over (I don't think you'll find any surprises).

Re: the one-or-many-PRs question, I'm a bit torn between "do it all for consistency" vs "cautious iterative changes"... but anything involving `@Extension` and/or `Descriptor` code always makes me uncomfortable (as I don't understand it nearly as well as I'd like) so in this case, caution wins.
i.e. I think that we should keep this PR as the _minimal_ set of changes required to update to Java 8, parent pom etc, and follow up with one (or more) PR(s) for other changes.

After this PR, I think that, as long as you're willing to do the necessary tests to ensure that "it's all good" then we can do it in as many or as few PRs as you want.  i.e. if you're driving it, it's up to you.  I'll just push the "merge" button when you say.

However, this is all contingent on (you) doing some manual tests (or you writing unit tests) to convince me that it not only works in isolation but also continues to work "as intended" when there are other plugins contributing other `@Extension` options.  If I have to do much testing myself then you'll have to wait a fair while before anything gets merged because, right now, I don't have much time for any of that.
i.e. it has to be tested before merger, and I'd like you to do that.

As far as testing goes, we don't want vSphere-specific launchers being suggested as an option for non-vsphere slaves, so that _either_ means a unit-test (which would be the preferred option) _or_ it means manual testing (typically by running up a test Jenkins with this plugin and a few others and checking that they're not interfering with each other in unexpected ways).

So, overall, I think we need to build up a list of `Descriptor`s this code change might affect, and define some test cases for how to test that it's working (a) as before, (b) without confusing other plugins and (c) without being confused by other plugins, and then make sure that those test cases are either unit-tested or manually-tested before we merge.
That, obviously, starts with the descriptor(s) affected by this PR...

TL;DR: Convince me that this PR is all fine and I'll merge it; I don't want to have to test it myself :stuck_out_tongue:

src/main/java/org/jenkinsci/plugins/vSphereCloudSlave.java
```diff
@@ -365,6 +367,14 @@ public void preLaunch(Computer c, TaskListener taskListener) throws IOException,
         }
     }
 
+    @Extension
+    public static class DescriptorVisibilityFilterImpl extends DescriptorVisibilityFilter {
+        @Override
+        public boolean filter(@CheckForNull Object context, @NonNull Descriptor descriptor) {
+            return !(descriptor instanceof vSphereCloudProvisionedSlave.DescriptorImpl);
```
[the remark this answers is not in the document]
casz: copy pasta 😆

src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
casz: you can ask IDEA to switch stream to foreach it will look exactly the same.

pom.xml
```diff
@@ -116,23 +112,4 @@
...
-            </plugin>
-            <plugin>
-                <groupId>org.codehaus.mojo</groupId>
-                <artifactId>findbugs-maven-plugin</artifactId>
-                <configuration>
-                    <skip>true</skip>
-                </configuration>
-            </plugin>
-        </plugins>
-    </build>
```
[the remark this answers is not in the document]
casz: the parent pom switched away from findbugs and started using spotbugs as findbugs project was inactive.

`spotbugs.skip` is a shorthand for the build plugin configuration.

By default in the new parent pom injected tests are always run.

src/main/java/org/jenkinsci/plugins/vSphereCloudSlave.java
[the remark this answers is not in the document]
casz:
```suggestion
            return !(descriptor instanceof DescriptorImpl);
```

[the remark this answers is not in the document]
pjdarton: I'd go with the former.  Where we've got
```
class C {
  class Inner { ... }
```
then code in class C that references class C.Inner should do so by saying `Inner` not `C.Inner`.

...but I'm still not entirely convinced that, even expressed like that, the code is doing what's needed, and the fact that this slipped through suggests that whatever testing you did, didn't pick this up.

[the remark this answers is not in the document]
casz: The only way to test it is through UI and I didn't test it since I just copy pasted ;) from one location which was the wrong thing to do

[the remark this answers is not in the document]
casz: It's been battled tested for the past year 😆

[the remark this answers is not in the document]
casz: Actually if you notice, this is the correct one implemented in `vSphereCloudProvisionedSlave` but since I copy pasted it from `vSphereCloudProvisionedSlave` to `vSphereCloudSlave`. Intellij decided to use `vSphereCloudProvisionedSlave.DescriptorImpl`

```
    @Extension
    public static class DescriptorVisibilityFilterImpl extends DescriptorVisibilityFilter {
        @Override
        public boolean filter(@CheckForNull Object context, @NonNull Descriptor descriptor) {
            return !(descriptor instanceof DescriptorImpl);
        }
    }
```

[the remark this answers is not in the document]
pjdarton: FYI Eclipse does that sort of thing too - decides that it knows better than to paste exactly the text you copied and instead tried to edit the code so it does exactly the same things as what you copied.  99% of the time, this isn't what I want...

...and, yes, to test it requires using the GUI.  Most of the `Descriptor` code is purely there to support the GUI.
I know that your primary use-case is JCasC, but this plugin can't abandon the 99.9% of other users who use the Jenkins configuration GUI to configure Jenkins, so that means that we _must_ test this before it gets merged.

src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
casz: Switched to foreach

src/main/java/org/jenkinsci/plugins/vSphereCloudSlave.java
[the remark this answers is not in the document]
casz: Fixed

casz:
👋 @pjdarton

pjdarton:
I'm not _deliberately_ ignoring you ... work's been a tad busy.
My remaining concern is one of testing - we need to ensure that none of these changes has an adverse affect on the ability to configure new vSphere slaves (either cloud or static), and you previously indicated that you've not tried that (that's how the unintended IDE changes went unnoticed) so, while I'm sure that the code changes don't break anything for your use-case, it might break things for a wider audience.

i.e. either you test it (see my previous comment), or you'll have to wait for me to test it.

casz:
I have tested it. We are using the updated code in production.

Again "not tested" part was in the attempt of splitting the PR in two #100 and #110 as you initially requested.

casz:
I'd like to consume the incremental build (I would need to add a commit to enable incremental repo publishing)

casz:
incremental allows me to do this in our production setup

```diff
diff --git a/dockerfile b/dockerfile
index b6022ed..26a139d 100644
--- a/dockerfile
+++ b/dockerfile
@@ -30,9 +30,7 @@ COPY --chown=jenkins plugins.txt $jenkins_refdir/

 # plugin installation layer, plugins are redownloaded everytime when plugins.txt is changed
 RUN mkdir -p $jenkins_refdir/plugins && \
-    install-plugins.sh < $jenkins_refdir/plugins.txt && \
-    curl -sSfL $artifactory/maven/org/jenkins-ci/plugins/vsphere-cloud/2.21-SNAPSHOT/vsphere-cloud-2.21-20191104.103446-1.hpi \
-    -o $jenkins_refdir/plugins/vsphere-cloud.jpi.override
+    install-plugins.sh < $jenkins_refdir/plugins.txt

 ARG CONTAINER_BUILD_TAG
 ENV CONTAINER_BUILD_TAG=$CONTAINER_BUILD_TAG
diff --git a/plugins.txt b/plugins.txt
index 40d1dd9..ddc27d7 100644
--- a/plugins.txt
+++ b/plugins.txt
@@ -132,8 +132,7 @@ token-macro:2.10
 trilead-api:1.0.5
 variant:1.3
 view-job-filters:2.1.1
# PR in progress: https://github.com/jenkinsci/vsphere-cloud-plugin/pull/100
-# vsphere-cloud:2.21-SNAPSHOT (private-d8485ff2-Joseph Petersen)
+vsphere-cloud:incrementals;org.jenkins-ci.plugins;2.21-rc447.075d17a3a216
 vstestrunner:1.0.8
 warnings-ng:7.1.0
 windows-azure-storage:1.1.2
```

pom.xml
```diff
@@ -54,10 +54,12 @@
         <connection>scm:git:git://github.com/jenkinsci/vsphere-cloud-plugin.git</connection>
         <developerConnection>scm:git:ssh://git@github.com/jenkinsci/vsphere-cloud-plugin.git</developerConnection>
         <url>https://github.com/jenkinsci/vsphere-cloud-plugin</url>
-        <tag>HEAD</tag>
+        <tag>${scmTag}</tag>
     </scm>
 
     <properties>
+        <revision>2.22</revision>
+        <changelist>-SNAPSHOT</changelist>
```
pjdarton: These two lines look like they're duplicating/conflicting-with the next two lines.

pjdarton:
If you hadn't snuck that `incrementalify` commit in, I'd've merged it just now ... but I think the pom is now slightly malformed with more than one value for the properties defined:
```
    <properties>
        <revision>2.22</revision>
        <changelist>-SNAPSHOT</changelist>
        <revision>2.21</revision>
        <changelist>-SNAPSHOT</changelist>
        <java.level>8</java.level>
        <jenkins.version>2.60.3</jenkins.version>
        <spotbugs.skip>true</spotbugs.skip>
    </properties>
```

Note: I would prefer incremental support to have been added as a separate PR so if it's just as easy to remove that commit from here and put it in a new PR then please do so.  If not, it'll have to be fixed here and merging this will have to wait until that's fixed.

casz:
I can't keep separating PRs... to do honest work 😭

pjdarton:
OK.  Are you happy for me to merge what we've got here as of https://github.com/jenkinsci/vsphere-cloud-plugin/pull/110/commits/25989eb6bd0b47f351aa0962868b6645c3546d4b ?

casz:
yes please. I then need to add a new PR... that I then need #100 to be rebased on 😭

src/main/java/org/jenkinsci/plugins/vsphere/VSphereConnectionConfig.java
danjasmin: With this change, I can no longer select any credentials since the dropdown list only contains “-none-“. I reverted the change to this method and the dropdown list is properly populated.
I would like to understand why the change was made and if the problem I am seeing is a bug or not. Any feedback you can supply would be appreciated.

pjdarton: Hmm.... my _guess_ is that you don't have sufficient rights on your Jenkins server to see all the credentials you're used to seeing.
The old code specified
```
                return new StandardListBoxModel().withEmptySelection().withMatching(
                    CREDENTIALS_MATCHER, CredentialsProvider.lookupCredentials(StandardCredentials.class,
                    instance, ACL.SYSTEM, getDomainRequirement(vsHost))
```
whereas the new code is missing the `ACL.SYSTEM` bit, so I would _guess_ that it's showing you what credentials you're allowed to see instead of all the credentials the system can see.
...but that's a guess - I'm not hugely familiar with how `StandardListBoxModel` is (or should be) used so I can't tell the difference between "good enough that it'll work in my own local environment" and "actually correct".
I can tell you that it "works for me" - if I go to "manage jenkins" -> "configure system", scroll down to "clouds" and then click on the drop-down list of credentials for a specific vSphere cloud then I see lots of credentials ... _but_ I'm logged as "as administrator" and I don't have any restrictions on how my credentials are used.  My assumption is that your setup differs, which is why it's not working for you.

If we can get to the bottom of this, I'd be quite happy to receive a PR that fixes it for you (as long as it doesn't break it for anyone else)...

casz: I believe for this to work with folder credentials it might need some tweaking

pjdarton: Ah, yes, folders make everything vastly more complicated, and they're not something I have any experience with.

If you are able to try out different options ... try using the new code but where it says `instance` put instead `containingFolderOrNull!=null?containingFolderOrNull:instance`
That should then provide the folder as the "context" for the credentials list instead of the Jenkins administrator-only list.

...and if that works then I suspect that we'll need similar changes elsewhere too...

albers: This change seems to be the cause of [JENKINS-61131](https://issues.jenkins-ci.org/browse/JENKINS-61131).
Can please a maintainer take a look at my proposed fix in #116?

pjdarton: Looked; merged.

@albers Given that you know enough on this subject to be able to code a fix ... what's your opinion on `containingFolderOrNull!=null?containingFolderOrNull:Jenkins.getInstance()` vs `Jenkins.getInstance()` ?
My own experience with credential lists is more "shooting in the dark" than actual informed opinion so if you know enough to suggest fixes you probably know more about how this _should_ be done than I do - I'd welcome any informed suggestions, especially if it'll let us get it right this time - trial and error isn't an approach I like to use...

Jetersen: ContainingFolderOrNull is not the proper way of doing it.

From jenkins you can ask for the context using `@AncestorInPath Item item`

https://github.com/jenkinsci/git-plugin/blob/c0e8144da63e3c8ccea7f00f7438b6e70aac9b00/src/main/java/hudson/plugins/git/UserRemoteConfig.java#L99-L121

pjdarton: While I am familiar (as in, I've seen it used before) with `@AncestorInPath Item item`, I'm less certain about `Tasks.getAuthenticationOf((Queue.Task) project)`; I suspect that the vSphere plugin is unlikely to encounter a `Queue.Task` and that this is more specific to the git plugin's use-case...

The existing vsphere-plugin code already obtains `@AncestorInPath AbstractFolder<?> containingFolderOrNull` and as `AbstractFolder` is `instanceof Item` I would expect `containingFolderOrNull` to be the same value we'd get if this code had requested `@AncestorInPath Item item` instead.
What I'm less certain of is exactly what to do with it...

PR #112, opened 2019-11-05
Add descriptor visibility filter

casz:


pjdarton:
My concern with these is that the visibility filter is saying "These must not be shown to anyone" whereas the truth is that they _should_ be an option for vSphere cloud-provisioned slaves (but _only_ vSphere cloud-provisioned slave nodes).
I suspect that the `@CheckForNull Object context` argument is intended to be used to allow visbility filters to say "we're only visible to vSphere cloud-provisioned slave nodes".

I think this would only ever become a problem if `vSphereCloudSlaveTemplate.DescriptorImpl.getRetentionStrategyDescriptors()` was re-coded to ask Jenkins to list all appropriate retention strategies instead of hard-coding that list tho'

If you agree with that, and you're happy for this to be merged, please add a PR description and I'll merge it.

pjdarton:
I suspect an end-of-line issue - the diff is now showing you've replaced the entire contents of both files...

casz:
:scream: I just used GitHub to merge a simple conflict 😭

pjdarton:
Of course, fixing the EOLs in #114 has meant this now has merge issues ... again :grinning:

casz:
i'll double check on the visibility but I recall them as being visible and selectable :)

pjdarton:
Thank you for your perseverance :+1:

pjdarton:
PS. It's the end of my working day now, but if you find any other issues, comment/raise-a-PR etc and I'll be back online in around 15 hours.

casz:
Just tested with the latest incremental build from master. Visibility filter works just fine: 
image

pjdarton:
For a cloud-provisioned slave, none of that should be editable anyway... They should be edited by configuring the template that defines how they're created, but that's (probably) a bug unrelated to your changes - one of many minor bits of tech-debt in this plugin code.

The important things are that:
1. When creating or editing a (vSphere slave) template, the vSphere retention strategies are available to be selected.  I think _that_ code has a hard-coded list of just the two vSphere options and doesn't pay any attention to visibility filters, so that should be unaffected.
1. When creating a new non-cloud vSphere slave, the vSphere retention strategies are available to be selected.
1. When creating any other kind of slave, the vSphere retention strategies are _not_ available.
1. When _editing_ a previously-created slave, the same retention strategies are listed as were available when it was created.

PR #113, opened 2019-11-05
fix value accessor

casz:
Fix accessor usage in CloudSelectorParameter as `.value` is (about to be) prohibited.


casz:
FYI this was the commit that changed the `.value` access https://github.com/jenkinsci/jenkins/commit/95db56e850f13650d3b16bec7a07f0abe20a931a#diff-ead88d327108535bd8df14b3e5d9d326R41

pjdarton:
Yup; I recognised it, saw it was obviously beneficial and _very_ unlikely to cause problems, so merged it as-is.

PR #114, opened 2019-11-05
fix line ending 😱

casz:
Fixed so line ending merge conflicts won't happen again.

by adding git attributes file this won't happen again.


casz:
thank goodness for `git add --renormalize`

pjdarton:
I was going to comment that "on my local machine, everything seems to be CR-LF end-of-line codes" ... but then, it's possible that this was because my local machine is Windows and, when I started on this plugin, I wouldn't've known about gitattributes and how important they are :grin:
Thanks for sorting this one - another bit of long-overdue-tech-debt resolved...

...I'll merge it once the automated build has checked that this hasn't upset anything.

pjdarton:
...I'm impatient - the automated test has got past the build step and is just waiting for a maven executor for the deploy (which IME can take a while).
I'll merge it anyway.

PR #118, opened 2020-08-18
Fix Jenkins 61702

pjdarton:
After 2.21, we (accidentally) lost the ability to manually create vSphere nodes from the UI.
See https://issues.jenkins-ci.org/browse/JENKINS-61702
Oops.

This is an attempt to solve that.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/vSphereCloudProvisionedSlave.java
+++ b/src/main/java/org/jenkinsci/plugins/vSphereCloudProvisionedSlave.java
@@ -100,89 +100,19 @@ public class vSphereCloudProvisionedSlave extends vSphereCloudSlave {
     }
 
     @Extension
-    public static class DescriptorVisibilityFilterImpl extends DescriptorVisibilityFilter {
-        @Override
-        public boolean filter(@CheckForNull Object context, @NonNull Descriptor descriptor) {
-            return !(descriptor instanceof DescriptorImpl);
-        }
-    }
-
-    @Extension
-    public static final class DescriptorImpl extends SlaveDescriptor {
-
-        public DescriptorImpl() {
-            load();
-        }
-
+    public static final class DescriptorImpl extends vSphereCloudSlave.DescriptorImpl {
         @Override
         public String getDisplayName() {
-            return "Slave created from a vSphere Cloud slave template";
+            return super.getDisplayName()+", auto-provisioned by Jenkins from cloud template";
         }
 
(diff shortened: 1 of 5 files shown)


pjdarton:
The original jenkinsci build failed because of infrastructure reasons (one AWS node died shortly after the build had successfully built, but before Jenkins had let it go, so the node death flagged the build as a failure).
Fortunately, it did yield a hpi file, https://ci.jenkins.io/job/Plugins/job/vsphere-cloud-plugin/job/PR-118/lastCompletedBuild/

To anyone watching this, it's a good idea to check the details of a "failed" build, as they're often unrelated to the code being built.

pjdarton:
This is now awaiting "someone other than me" to test that it fixes the issue it's meant to.
I've tried it within my own environment and I've checked that it doesn't seem to have broken anything, and my own (shallow) testing suggests it might've fixed it, but as I wasn't suffering from the issue reported I can't confirm/deny it's 100% sorted now.

However, once we've got positive confirmation (either here or on the [JIRA issue](https://issues.jenkins-ci.org/browse/JENKINS-61702)) from someone who was hitting this problem that they've tried it and it fixes the issue, I'll be happy to merge it it and it can then go into the next release.

pjdarton:
@jetersen Can you (also) test that this PR doesn't break JCasC?
This PR is reworking some of the code modified in #110 and so there's the potential that by fixing this issue I'll end up breaking JCasC.

src/main/java/org/jenkinsci/plugins/vSphereCloudProvisionedSlave.java
```diff
@@ -100,89 +100,19 @@ public void preLaunch(Computer c, TaskListener taskListener) throws IOException,
...
+            return super.getDisplayName()+", auto-provisioned by Jenkins from cloud template";
         }
 
         @Override
         public boolean isInstantiable() {
+            /*
+             * This type of agent can't be directly created by the user through the UI.
+             * The user defines a vSphere agent template and _that_ then creates these "on demand".
+             */
             return false;
```
jetersen, line 115: I think we are fine as long as this is kept in. This is meant to prevent JCasC from removing nodes provisioned from this class.

pjdarton:
Feedback from [JENKINS-61702](https://issues.jenkins-ci.org/browse/JENKINS-61702) is that this PR fixes the bug.
That means we're all clear to merge.

PR #124, opened 2020-11-30
Fixes [JENKINS-64335]

balleman-tt:
* Do not set slaveIsDisconnecting for temporarily offline nodes


pjdarton:
Thank you for tracking this down & raising this PR 👍 

That code has _way_ too much complexity in it for my liking (the joys of old old code!) and, yes, I agree that it looks like commit 7ea9d7ae1d57b6f38456ddd6b92873f651bf317b did change the order; you did a good bit of detective work there 😉 

Code changes here look good ... mind you, the code changes in #117 looked perfectly reasonable too so 🤞

PR #125, opened 2020-12-19
Java 11 compatibility and refresh dependencies

timja:
Add bom so that dependencies are tested against each other, known to be compatible and guaranteed not to cause dependency conflicts with enforcer

exclusion added on Lombok because of impropery packaged dependency that breaks compilation on java 11

updated to a recent core to support building with java 11, bom, Jenkins core bom, to test against a more supported version of Jenkins that’s more realistic to what users use, version chosen by following https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,7 +61,7 @@
         <revision>2.25</revision>
         <changelist>-SNAPSHOT</changelist>
         <java.level>8</java.level>
-        <jenkins.version>2.235.5</jenkins.version>
+        <jenkins.version>2.190.1</jenkins.version>
         <spotbugs.skip>true</spotbugs.skip>
     </properties>
 
@@ -69,8 +69,8 @@
         <dependencies>
             <dependency>
                 <groupId>io.jenkins.tools.bom</groupId>
-                <artifactId>bom-2.235.x</artifactId>
-                <version>19</version>
+                <artifactId>bom-2.190.x</artifactId>
+                <version>16</version>
                 <scope>import</scope>
                 <type>pom</type>
             </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.51</version>
+        <version>4.14</version>
     </parent>
 
     <artifactId>vsphere-cloud</artifactId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -61,27 +61,50 @@
         <revision>2.25</revision>
         <changelist>-SNAPSHOT</changelist>
         <java.level>8</java.level>
-        <jenkins.version>2.60.3</jenkins.version>
+        <jenkins.version>2.235.5</jenkins.version>
```
pjdarton, line 64: Does it _have_ to be this recent?  What's the justification?

I prefer the minimum version of Jenkins necessary to be as old as possible, but 2.235.5 is only 2020-08-17 and I'd prefer at least a year old, otherwise we're forcing folks to update who may not want to (it happens; being on the latest isn't for everyone).
So, I'd prefer it if we can rebase on 2.190.3 (or earlier) instead - that's (just) over a year old.

pjdarton (changes requested):
While an overhaul (by someone who clearly understands Jenkins dependencies etc more than I do) is most welcome, making the minimum Jenkins version more recent than a year old is less so.

So.... how far back in the past can we push the minimum without losing too many of the benefits?

pom.xml
timja: We can do 2.190 if you want, I based it off of this: https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/

The reason to upgrade was the plugin didn't compile on my machine without updating it, also needs to be at least old enough to include the Jenkins bom 2.170 something I think so that testing against super new versions doesn't require hacking in the pom

pjdarton:
Can you provide a description please?  What's the intent behind these changes?  Why is this necessary etc?

timja:
The background is for this: https://github.com/jenkinsci/vsphere-cloud-plugin/pull/126

pom.xml
pjdarton: OK, I've read that guidance (just now) and I'd opt for the oldest possible, so 2.190.1 in this case.
I also checked the installed [stats for this plugin](https://stats.jenkins.io/pluginversions/vsphere-cloud.html) and everyone who's on the latest plugin version is on 2.190.1 or later, so making 2.190.1 the minimum Jenkins version this shouldn't prevent anyone using the current latest version of this plugin from upgrading when I do a new release.

...it's also weird that those stats show a plugin version 3.0.3.0 and 3.1.1.0 because there's never (officially) been a 3.x version.  I've no idea where thost stats are coming from; maybe those are unofficial custom builds? ... but I'm not going to lose any sleep over that weirdness - our "lets not upset people" target audience is those who are using the latest version of this plugin.

pjdarton (approved):

pom.xml
timja: > maybe those are unofficial custom builds

yes exactly, would be nice if they put their org name in there or snapshot as those are filtered out but not much we can do if they publish them with legit looking version

PR #126, opened 2020-12-19
JENKINS-64357 Fix plugin on Jenkins >2.263

timja:
See https://issues.jenkins.io/browse/JENKINS-64357

and https://www.jenkins.io/doc/developer/views/table-to-div-migration/

Cleanup needed for me to compile this plugin (running java 11) is in https://github.com/jenkinsci/vsphere-cloud-plugin/pull/125

@pjdarton

### DIFF
--- a/src/main/resources/lib/vsphere/blockWrapper.jelly
+++ b/src/main/resources/lib/vsphere/blockWrapper.jelly
@@ -1,0 +1,16 @@
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:d="jelly:define">
+    <!-- TODO remove and switch to div after baseline is 2.264 or newer -->
+    <j:choose>
+        <j:when test="${divBasedFormLayout}">
+            <div>
+                <d:invokeBody/>
+            </div>
+        </j:when>
+        <j:otherwise>
+            <table style="width:100%">
+                <d:invokeBody/>
+            </table>
+        </j:otherwise>
+    </j:choose>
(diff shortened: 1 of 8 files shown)


src/main/resources/lib/vsphere/blockWrapper.jelly
```diff
@@ -0,0 +1,16 @@
...
+                <d:invokeBody/>
+            </div>
+        </j:when>
+        <j:otherwise>
+            <table style="width:100%">
+                <d:invokeBody/>
+            </table>
+        </j:otherwise>
+    </j:choose>
+</j:jelly>
```
pjdarton, line 16: I presume that this is the core of the fix?
i.e. pre 2.264 it uses `table` and post it uses `div`?

...and does it do this automatically?  Is that what `test="${divBasedFormLayout}"` achieves?

pjdarton (commented):
I think we need to resolve (PR 125)[https://github.com/jenkinsci/vsphere-cloud-plugin/pull/125#discussion_r546949281] before this one.
...but the only contentious stuff here (aside from #125) is down to my lack of understanding and request for explanations...

src/main/resources/lib/vsphere/blockWrapper.jelly
timja: Yes it’s a property on set on >=2.264

timja:
@pjdarton could this be released please?

pjdarton:
Done.

PR #127, opened 2021-02-17
Add nodes pre-provisioning

korablin:
In order to have instances at the ready at all times and
thus speed up CI runs because instance provisioning is done
upfront, add a mechanism and the related parameter to configure
a minimum number of instances that must be ready at all times.

Also add a mechanism to reconcile total number of nodes
based on template/cloud capacity

There is similar functionality presented in openstack cloud plugin:
https://github.com/jenkinsci/openstack-cloud-plugin/blob/master/plugin/src/main/java/jenkins/plugins/openstack/compute/JCloudsPreCreationThread.java

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/vSphereCloud.java
+++ b/src/main/java/org/jenkinsci/plugins/vSphereCloud.java
@@ -202,6 +202,9 @@ public class vSphereCloud extends Cloud {
             }
         }
     }
+    public ConcurrentHashMap<String, String> getCurrentOnlineSlaves() {
+        return currentOnline;
+    }
 
     public int getMaxOnlineSlaves() {
         return maxOnlineSlaves;
@@ -215,7 +218,7 @@ public class vSphereCloud extends Cloud {
         return this.templates;
     }
 
-    private vSphereCloudSlaveTemplate getTemplateForVM(final String vmName) {
+    public vSphereCloudSlaveTemplate getTemplateForVM(final String vmName) {
         if (this.templates == null || vmName == null)
             return null;
         for (final vSphereCloudSlaveTemplate t : this.templates) {
--- a/src/main/java/org/jenkinsci/plugins/vsphere/VSphereNodeReconcileWork.java
+++ b/src/main/java/org/jenkinsci/plugins/vsphere/VSphereNodeReconcileWork.java
@@ -1,0 +1,119 @@
+package org.jenkinsci.plugins.vsphere;
+
+import hudson.Extension;
+import hudson.Functions;
+import hudson.model.TaskListener;
+import hudson.model.AsyncPeriodicWork;
+import hudson.slaves.AbstractCloudComputer;
+import hudson.slaves.AbstractCloudSlave;
+import hudson.slaves.Cloud;
+import hudson.model.Label;
+import jenkins.model.Jenkins;
+import org.jenkinsci.plugins.vSphereCloud;
+import org.jenkinsci.plugins.vSphereCloudSlave;
+import org.jenkinsci.plugins.vSphereCloudSlaveComputer;
+import org.jenkinsci.plugins.vSphereCloudSlaveTemplate;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+
+/**
+ * A {@link AsyncPeriodicWork} that reconciles nodes to meet template values.
+ * <p>
+ * The async work will check the number of deployed nodes and provision (or
+ * delete) additional ones to meet template values. The check is happening every
+ * 2 minutes.
+ */
+@Extension
+@Restricted(NoExternalUse.class)
+public final class VSphereNodeReconcileWork extends AsyncPeriodicWork {
+    private static final Logger LOGGER = Logger.getLogger(VSphereNodeReconcileWork.class.getName());
+
+    public VSphereNodeReconcileWork() {
+        super("Vsphere nodes reconciliation");
+    }
+
+    @Override
+    public long getRecurrencePeriod() {
+        return Functions.getIsUnitTest() ? Long.MAX_VALUE : MIN * 2;
+    }
+
+    @Override
+    public void execute(TaskListener listener) {
+        for (Cloud cloud : Jenkins.getActiveInstance().clouds) {
+            if (!(cloud instanceof vSphereCloud)) continue;
+            for (vSphereCloudSlaveTemplate template : ((vSphereCloud) cloud).getTemplates()) {
+                String templateLabel = template.getLabelString();
+                Label label = Label.get(templateLabel);
+
+                int instancesMin = template.getInstancesMin();
+                List<vSphereCloudSlaveComputer> idleNodes = template.getIdleNodes();
+                List<vSphereCloudSlaveComputer> runningNodes = template.getOnlineNodes();
+                List<vSphereCloudSlaveComputer> reusableBusyNodes = template.getBusyReusableNodes();
+                // Get max number of nodes that could be provisioned
+                int globalMaxNodes = ((vSphereCloud) cloud).getInstanceCap();
+                int templateMaxNodes = template.getTemplateInstanceCap();
+                int maxNodes = Math.min(globalMaxNodes, templateMaxNodes);
+
+                // if maxNumber is lower than instancesMin, we have to ignore instancesMin
+                int toProvision = Math.min(instancesMin - (reusableBusyNodes.size() + idleNodes.size()),
+                        maxNodes - runningNodes.size());
+                if (toProvision > 0) {
+                    // provision desired number of nodes for this label
+                    LOGGER.log(Level.INFO, "Pre-creating {0} instance(s) for template {1} in cloud {3}",
+                            new Object[] { toProvision, templateLabel, cloud.name });
+                    try {
+                        cloud.provision(label, toProvision);
+                    } catch (Throwable ex) {
+                        LOGGER.log(Level.SEVERE, "Failed to pre-create instance from template {0}. Exception: {1}",
+                                new Object[] { templateLabel, ex });
+                    }
+                } else if (toProvision < 0) {
+                    int toDelete = Math.min(idleNodes.size(), Math.abs(toProvision));
+                    for (int i = 0; i < toDelete; i++) {
+                        AbstractCloudSlave node = idleNodes.get(i).getNode();
+                        if (node == null) continue;
+                        LOGGER.log(Level.INFO, "Found excessive instance. Terminating {0} node {1}.",
+                                new Object[] { idleNodes.get(i).getName(), node });
+                        try {
+                            node.terminate();
+                        } catch (InterruptedException | IOException e) {
+                            LOGGER.log(Level.WARNING, e.getMessage());
+                            // try to delete it later
+                            continue;
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    /**
+     * Should a node be retained to meet the minimum instances constraint?
+     */
+    @SuppressWarnings("rawtypes")
+    public static boolean shouldNodeBeRetained(AbstractCloudComputer c) {
+        // Checks only idle nodes
+        vSphereCloudSlave node = (vSphereCloudSlave) c.getNode();
+        if (node == null) return false;
+        vSphereCloudSlaveTemplate nodeTemplate = node.getTemplate();
+        // nodeTemplate might be null if the template was manually deleted from a cloud
+        if (nodeTemplate == null) return false;
+        int instancesMin = nodeTemplate.getInstancesMin();
+        if (instancesMin > 0) {
+            int maxNodes = Math.min(nodeTemplate.getTemplateInstanceCap(), nodeTemplate.getParent().getInstanceCap());
+            int runningNodesTotal = nodeTemplate.getOnlineNodes().size();
+            int idleNodesTotal = nodeTemplate.getIdleNodes().size();
+            int reusableBusyNodesTotal = nodeTemplate.getBusyReusableNodes().size();
+            if ((instancesMin >= (idleNodesTotal + reusableBusyNodesTotal - 1)) && (runningNodesTotal <= maxNodes)) {
+                return true;
+            }
+        }
+        return false;
+    }
+}
--- a/src/main/resources/org/jenkinsci/plugins/vSphereCloudSlaveTemplate/config.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/vSphereCloudSlaveTemplate/config.jelly
@@ -54,6 +54,10 @@
                 <f:textbox clazz="required number" default="0"/>
             </f:entry>
 
+            <f:entry title="${%Instances Min}" field="instancesMin" description="The number of instances to be pre-provisioned.">
+                <f:textbox clazz="required number" default="0"/>
+            </f:entry>
+
             <f:entry title="${%# of Executors}" field="numberOfExecutors">
                 <f:textbox clazz="required positive-number" default="1"/>
             </f:entry>
--- a/src/main/java/org/jenkinsci/plugins/vSphereCloudSlave.java
+++ b/src/main/java/org/jenkinsci/plugins/vSphereCloudSlave.java
@@ -133,6 +133,11 @@ public class vSphereCloudSlave extends AbstractCloudSlave {
         return LimitedTestRunCount;
     }
 
+    public vSphereCloudSlaveTemplate getTemplate() {
+        vSphereCloud cloud = findOurVsInstance();
+        return cloud.getTemplateForVM(getVmName());
+    }
+
     public boolean isLaunchSupportForced() {
         return ((vSphereCloudLauncher) getLauncher()).getOverrideLaunchSupported() == Boolean.TRUE;
     }
--- a/src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveComputer.java
+++ b/src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveComputer.java
@@ -2,7 +2,10 @@ package org.jenkinsci.plugins;
 
 import java.io.PrintWriter;
 import java.io.StringWriter;
+import java.util.ArrayList;
+import java.util.List;
 
+import javax.annotation.Nonnull;
 import org.jenkinsci.plugins.vsphere.tools.VSphere;
 
 import com.vmware.vim25.VirtualHardware;
@@ -14,8 +17,10 @@ import com.vmware.vim25.VirtualMachineToolsStatus;
 import com.vmware.vim25.mo.ManagedEntity;
 import com.vmware.vim25.mo.VirtualMachine;
 
+import hudson.model.Computer;
 import hudson.slaves.AbstractCloudComputer;
 import hudson.slaves.AbstractCloudSlave;
+import jenkins.model.Jenkins;
 
 public class vSphereCloudSlaveComputer extends AbstractCloudComputer {
     private final vSphereCloudSlave vSlave;
@@ -75,6 +80,18 @@ public class vSphereCloudSlaveComputer extends AbstractCloudComputer {
 
     public String getVmInformationError() {
         return getVMInformation().errorEncounteredWhenDataWasRead;
+    }
+
+    /**
+     * Get all computers.
+     */
+    public static @Nonnull List<vSphereCloudSlaveComputer> getAll() {
+        ArrayList<vSphereCloudSlaveComputer> out = new ArrayList<>();
+        for (final Computer c : Jenkins.get().getComputers()) {
+            if (!(c instanceof vSphereCloudSlaveComputer)) continue;
+            out.add((vSphereCloudSlaveComputer) c);
+        }
+        return out;
     }
 
     /** 10 seconds */
--- a/src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
+++ b/src/main/java/org/jenkinsci/plugins/vSphereCloudSlaveTemplate.java
@@ -49,7 +49,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import java.util.stream.Collectors;
 
 import javax.annotation.Nonnull;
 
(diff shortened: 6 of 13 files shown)


pjdarton:
Thanks for writing this - it's functionality that I approve of 👍 
I'm familiar with the openstack plugin and I like that feature 😉 

However, I'm not currently in a position to review this in detail right now, so I'll stick to general points:

`public` stuff: while this plugin code has loads more things public than it should, I don't want the problem made any worse. Please ensure that all externally-visible API additions are marked `@Restricted(NoExternalUse.class)` ... and use package access instead of public unless public is absolutely necessary.

WebUI: use help pages rather than description attributes. I find that description text bloats the (already long) page more than having a help icon on the right ... and this functionality could do with a big explanation of what it does so the help text would be a perfect place for that.

VSphereCloudRetentionStrategy.check might be better if it did it's thing and then, if the node isn't needed, delegated to super.check rather than copy/paste what super.check currently does ... or else, why not do what RunOnce does and call done(c)?  If there is a good reason why it has to be different, please comment why.

Unit tests would be appreciated; I know we can't test against a real vSphere instance, but we can mock things and check the code is requesting the changes we expect.
(And, yes, I admit that most of this code has no unit tests; I'd like to change that by having new functionality have better QA as this plugin is enough of a maintenance-nightmare as it is)

korablin:
Hi @pjdarton! 
Thanks for your fast and informative response. I have addressed your comments with the second commit except for the unit tests. I wouldn't mind to add them, but I don't see an easy way of mocking things that are needed to test this PR.
I would love to cover this code with tests once upstream will provide some guide/mechanism for easy testing.

Could you take a look once again at the code and see if something is still too 'public'? I am not very good at Java, so I could have missed a thing.

Thanks a lot in advance!

Regards,
Nikolai

src/main/resources/org/jenkinsci/plugins/vSphereCloudSlaveTemplate/config.jelly
```diff
@@ -54,6 +54,10 @@
                 <f:textbox clazz="required number" default="0"/>
             </f:entry>
 
+            <f:entry title="${%Instances Min}" field="instancesMin" description="The number of instances to be pre-provisioned.">
```
pjdarton, line 57: We don't need/want a description when we have help text.

src/main/resources/org/jenkinsci/plugins/vSphereCloudSlaveTemplate/help-instancesMin.html
```diff
@@ -0,0 +1,5 @@
+<div>
+    Number of instances to be provisioned beforehand.<br>
```
pjdarton, line 2: Can we have this of the form:

one line summary
<p>
Multiple sentences explaining what this does.
Possibly using <br> where appropriate.

If appreciate it if you can expand the description.
Tell folks what a value of 0 does.
Tell folks what a value of 1 does, and what happens when that one agent stays busy for ages.

... and my *guess* is that if you set it to the same value as the max instances then it'll try to keep that number provisioned regardless of Jenkins workload, replacing any VMs that close e.g. due to "run once" retention strategies... but if that's the case, please state this.
(Nobody ever complains about having too much help text, but too little often results in folks raising an issue asking WTF stuff does, and I'd prefer to avoid that)

pjdarton (commented):
Code looks good. All the newly-added public stuff now has `@Restricted` on it which is good.

WebUI resource files need tweaking.

Disclaimer: I've not tried this out myself yet; I'm just looking at the code changes on GitHub. I will (eventually) get around to trying it at work (and if it works, I'll make use of it right away!) but I'm not going to merge it until I've seen it working myself "for real".

pjdarton:
Note: if the Jenkins test fails and you don't think it's your fault, you should be able to get GitHub to re-run the test.
(and if GitHub doesn't let you do it easily, you can force it by closing the PR, waiting 10 minutes, then re-opening the same PR)

korablin:
Hey @pjdarton! Glad you are thrilled to get this functionality added. I agree that more docs is better than less.
What do you think about this one?
Screenshot 2021-02-19 at 17 56 35

korablin:
Screenshot 2021-02-19 at 18 48 56

korablin:
Screenshot 2021-02-19 at 21 41 38

korablin:
Hey @pjdarton! Hope you are doing well, mate.
Could you take a look at my last commit?

pjdarton (changes requested):
I think `Boolean` vs `boolean` is a simple coding error and easily fixed.

I suspect that many new methods are more visible than they need to be.  `public` should be the last-resort; default visibility is package-level and as most of the plugin's code is all in the same package that's often all that's needed. `protected` is generally only used to give access to subclasses.
...but we do need to ensure that all new non-private code is `@Restricted(NoExternalUse.class)` unless it's specifically part of the public API (so we don't restrict access to the new getter & setter fot the new field, but all the internal workings should be restricted)

I'm concerned that the "retain VM" functionality will actually defeat / prevent-from-working existing retention-strategy functionality that kills off VMs when they're no longer useful.
It may be better to have this functionality create new VMs to fulfil the instanceMin requirement but allow them to time out normally, instructing users 9in the documentation) that if they don't want instances to time out quickly like that then they'll need to increase the retention time.
The existing "if instancesPresent < instanceMin then ignore the retention strategy and keep it" logic check seems to be _too_ simplistic.

Dealing with the asynchronicity of retention-strategy checks and the instanceMin provisioning background task _appears_ to be improperly handled.  I suspect that you'll need some more complicated state handling for this to work properly.
Consider having the periodic task evaluate which instances ought to be kept, and in what priority order, and have the retention strategies then act on that evaluation result "when they get called" ... or (and I suspect this will be the much simpler option) have the periodic task _only_ be responsible for creating new instances if the minimum isn't met yet and let the retention strategies take care of disposing of them.

korablin:
@pjdarton I have killed `shouldSlaveBeRetained()` function and removed changes that I did to ret policies, so they their work is not anymore affected by this change.
Initially, I added those changes cause I wanted to make this behaviour consistent with OpenStack plugin, but guess we could iterate on this later if people will be complaining.

Regarding your question on using package-level visibility: I don't think it is possible.
Methods that reconcile scripts uses are located in `package org.jenkinsci.plugins`, but the script itself is in `package org.jenkinsci.plugins.vsphere` which appear to me as different packages.
Correct me if I am wrong. Also I am not sure if this the right place for this script.

korablin:
I am sorry for the belated response.
It would be nice to communicate quickly. I am willing to address your future comments asap.

src/main/java/org/jenkinsci/plugins/vsphere/VSphereNodeReconcileWork.java
```diff
@@ -0,0 +1,91 @@
...
+                int instancesMin = template.getInstancesMin();
+                List<vSphereCloudSlaveComputer> idleNodes = template.getIdleNodes();
+                List<vSphereCloudSlaveComputer> runningNodes = template.getOnlineNodes();
+                // Get max number of nodes that could be provisioned
+                int globalMaxNodes = ((vSphereCloud) cloud).getInstanceCap();
+                int templateMaxNodes = template.getTemplateInstanceCap();
+                int maxNodes = Math.min(globalMaxNodes, templateMaxNodes);
+
+                // if maxNumber is lower than instancesMin, we have to ignore instancesMin
+                int toProvision = Math.min(instancesMin - idleNodes.size(), maxNodes - runningNodes.size());
```
pjdarton, line 61: I'm not convinced that it's as simple as this as there's a delay between when a provisioning is started and when Jenkins is given a new agent and hence creates a new `vSphereCloudSlaveComputer` instance (and again when it's time to delete a VM too).

If you check the logic in vSphereCloud.provision(...) you'll see that it calls `ensureLists()` and then interrogates its `templateState` data to count how many VMs exist.
Some of those VMs will be in a state where the vSphere service has been asked to create a new VM but Jenkins won't know it exists yet; some of them will be fully created VMs that aren't online yet; some will be online and either busy or idle; some will be in the process of being deleted.
That `templateState` data structure is the "one true source of information" that we have (in memory) of what VMs we've asked for, what we've got, and what we've asked to delete.

The logic in use here only considers VMs that are fully created and online (and the subset of those which are idle); it does not consider those "not yet online" or "those being deleted".

I think that what we need here is to use the same logic (or at least be consistent with it) as used in vSphereCloud so that VMs which are being created are counted against the total present in the cloud, otherwise we could easily overshoot the desired `instancesMin` value (e.g. if VMs take so long to create that `public void execute(TaskListener listener)` gets called again before instances it's already requested have "arrived".
IMO this "pre-create" functionality will be most valued by folks whose vSphere infrastructure is slow to create new VMs, so this code _must_ cope with the infrastructure being slow.

src/main/java/org/jenkinsci/plugins/vsphere/VSphereNodeReconcileWork.java
```diff
@@ -0,0 +1,91 @@
...
+                    // provision desired number of nodes for this label
+                    LOGGER.log(Level.INFO, "Pre-creating {0} instance(s) for template {1} in cloud {3}",
+                            new Object[] { toProvision, templateLabel, cloud.name });
+                    try {
+                        cloud.provision(label, toProvision);
+                    } catch (Throwable ex) {
+                        LOGGER.log(Level.SEVERE, "Failed to pre-create instance from template {0}. Exception: {1}",
+                                new Object[] { templateLabel, ex });
+                    }
+                } else if (toProvision < 0) {
```
pjdarton, line 72: I am not sure we need this kind of logic here - the retention policy should allow unused and unnecessary nodes to be quietly disposed of.
I'm open to discussion on this, but e.g. it isn't deemed necessary in the OpenStack plugin and I suspect that its implementation works fine.

src/main/java/org/jenkinsci/plugins/vsphere/VSphereNodeReconcileWork.java
```diff
@@ -0,0 +1,91 @@
...
+                    int toDelete = Math.min(idleNodes.size(), Math.abs(toProvision));
+                    for (int i = 0; i < toDelete; i++) {
+                        AbstractCloudSlave node = idleNodes.get(i).getNode();
```
pjdarton, line 75: There's a potential race condition here:  By the time we get here, agents may not be idle anymore (and may not even exist within Jenkins anymore).
(Jenkins is _heavily_ multi-threaded - loads of things are happening all at once)

src/main/java/org/jenkinsci/plugins/vsphere/VSphereNodeReconcileWork.java
```diff
@@ -0,0 +1,91 @@
...
+                int maxNodes = Math.min(globalMaxNodes, templateMaxNodes);
+
+                // if maxNumber is lower than instancesMin, we have to ignore instancesMin
+                int toProvision = Math.min(instancesMin - idleNodes.size(), maxNodes - runningNodes.size());
+                if (toProvision > 0) {
+                    // provision desired number of nodes for this label
+                    LOGGER.log(Level.INFO, "Pre-creating {0} instance(s) for template {1} in cloud {3}",
+                            new Object[] { toProvision, templateLabel, cloud.name });
+                    try {
+                        cloud.provision(label, toProvision);
```
pjdarton, line 67: The problem here is that `label` isn't neccessarily unique to one template.

What we want here is to provision an instance for a specific template, rather than leaving it to the `provision` method to select a template that matches the specified label.
(FYI the `provision` method will distribute VMs over all templates matching the specified label if those templates have instance caps on them)

e.g. where I work, we have set of GPU-enabled servers on our vSphere cluster, each of which is capable of running N VMs, so we have one template per server with "max instances" = N, but aside from a specification to limit the template to a specific server, the templates are identical.  This means that if we're empty and then get a request for 5 instances that can run on either or two servers, we'll run 3 on one server and 2 on the other.
...although, in this example, all 5 would take time in arriving (as our vSphere cluster takes a while to clone a new VM from a vSphere template) and then take even longer before they're online (as the VMs run Windows and this takes a while to boot up).

If you compare what you're doing here with what was done on the [OpenStack PR](https://github.com/jenkinsci/openstack-cloud-plugin/pull/205/files) you'll see that the provisioning process in the cloud class was refactored to expose a "provision a new agent from this template" method; I'd suggest that we're going to need a similar thing here, otherwise this method will confuse templates and labels.

pjdarton (changes requested):

korablin:
@pjdarton  thanks for another iteration on this PR. I appreciate it.
So far, I pushed the minor changes fixing level of visibility (those things used to be public due to earlier functionality that I've killed).

Regarding the major things:
   - Tracking state of a request i.e. using templateState when deploying a VM + provisioning templates instead of labels
This sounds right to me. Could you give a hint on how do you envision implementation for this?
Should it be a method similar to `provision` in vSphereCloud that takes `template` and number of instances as arguments and then operates with `CloudProvisioningRecord` ? 

I would love to hear your opinion before I will start implementing it.


Also, current combination of provisioning things the simple way (as I do now) + deletion of redundant nodes within this code, kind of makes sure that we meet the instancesMin number, but of course there is a risk creating excessive VMs that will be killed on the second run of the periodic job.
^^ I am mentioning this just in case, if we would consider merging the code as is and iterate on it later with further improvements - that shouldn't blow up  your Jenkins :)

pjdarton:
Re: hint on how do you envision the implementation

A cloud plugin's method `public Collection<PlannedNode> provision(final Label label, int excessWorkload)` is how Jenkins asks the cloud-plugin to provide more nodes; this method choses a template to use and then creates a new node (and then repeats until enough executors are in progress, finally returning a set of "planned nodes" to Jenkins).
Within that code, it maintains the internal records (`templateState.pruneUnwantedRecords()`), ensures it doesn't exceed the cloud capacity (and returns `Collections.emptySet()` if so) before getting on with deciding which template(s) match the label Jenkins is asking for and then creating new VMs from the template it decided on.

This PR is adding a new method that'll provide more nodes; it'll need to iterate over all the templates checking that there's enough <existing idle nodes> + <nodes under construction> to satisfy the template's instancesMin field.


In this new case, we don't want to call `List<vSphereCloudSlaveTemplate> templates = getTemplates(label)` as, instead, we already know exactly which template we want (as it's iterating over all of them) ... but we could just make a list containing just the template we want before going on to call `templateState.calculateProvisionableTemplates(templates)` etc.

So, I think that what we do is

1. Refactor the existing vSphereCloud `provision` method so that _some_ of what it does is split out into separate methods (that our new code can call).  [This bit](https://github.com/jenkinsci/vsphere-cloud-plugin/pull/127/files#diff-4f085b36000ec240357f1171a2a163cad43c64c3e1bafacc8a60804007271cbdR374-R379) looks promising, but there may be others too.
2. Define a new `provisionInstanceMin` method in vSphereCloud that'll do `synchronized (this) { ensureLists(); }` and then, within a `synchronized (templateState)` block, will iterate over each of the `CloudProvisioningRecord`s in `templateState`, compare the `instanceMin` field's value with the number of `CloudProvisioningRecord.size()` (as that'll include all VMs, not just active nodes) minus the number of _active_ nodes for this template and provision some new ones if more are needed (and there's space for them!).  There's a fair amount of decision making etc that'll be identical to that done in the original `provision` method.
3. I would suggest you remove the existing code that'll dispose of unwanted instances - disposal is hard to do safely (e.g. as shown in [JENKINS-55066](https://issues.jenkins.io/browse/JENKINS-55066) and similar) and there are existing mechanisms that will slowly dispose of unwanted VMs safely.
4. have the periodic code loop over the vSphereClouds and, for each of those, call the new provisionInstanceMin method.
5. Finally, compare the new `provisionInstanceMin` method with the old vSphereCloud `provision` method and check again for common code that ought to be refactored into its own method that's then used by both.

...and be careful not to change the behaviour of the original `provision` method - no matter how much the code is rearranged, the functionality of that method needs to remain unchanged.

korablin:
@pjdarton It's been a while... I am sorry again. Ever-changing priorities are running plans sometimes.
Hopefully, with the last commit that refactors the whole thing for the better way, we will be able to merge it soon :)
Thanks for all of your comments! I have tried my best to address them and things look much simpler now.

I am willing to address further comments asap. 
Thank you @pjdarton in advance!

src/main/java/org/jenkinsci/plugins/vSphereCloud.java
```diff
@@ -352,22 +352,15 @@ private Integer calculateMaxAdditionalSlavesPermitted() {
...
                         + numberOfvSphereCloudSlaveExecutors + " executors), templates available are " + whatWeCouldUse);
                 while (excessWorkloadSoFar > 0) {
-                    if (maxSlavesToProvisionBeforeCloudCapHit != null) {
-                        final int intValue = maxSlavesToProvisionBeforeCloudCapHit.intValue();
-                        if (intValue <= 0) {
-                            break; // out of capacity due to cloud instance cap
-                        }
-                        maxSlavesToProvisionBeforeCloudCapHit = Integer.valueOf(intValue - 1);
-                    }
+                    if (!cloudHasCapacity()) break;
```
korablin, line 363: I am not subtracting 1 here, since `maxSlavesToProvisionBeforeCloudCapHit` is re-calculating on every call to `cloudHasCapacity()`

src/main/java/org/jenkinsci/plugins/vsphere/tools/CloudProvisioningAlgorithm.java
```diff
@@ -81,6 +81,23 @@ public static String findUnusedName(CloudProvisioningRecord record) {
...
+     * @param record
+     *            Our record regarding the template the agent will be created
+     *            from.
+     * @return A number of nodes to be provisioned.
+     */
+    public static int shouldPreProvisionNodes(CloudProvisioningRecord record) {
+        int provisionedNodes = record.getCurrentlyProvisioned().size() + record.getCurrentlyPlanned().size();
+        int requiredPreProvisionedNodes = record.getTemplate().getInstancesMin();
+        return requiredPreProvisionedNodes - provisionedNodes;
+    }
```
pjdarton, line 99: We're going to need a unit test for this new method.

src/main/java/org/jenkinsci/plugins/vsphere/VSpherePreProvisonWork.java
```diff
@@ -0,0 +1,52 @@
...
+    @Override
+    public long getRecurrencePeriod() {
+        return Functions.getIsUnitTest() ? Long.MAX_VALUE : MIN * 2;
+    }
+
+    @Override
+    public void execute(TaskListener listener) {
+        for (Cloud cloud : Jenkins.getActiveInstance().clouds) {
+            if (!(cloud instanceof vSphereCloud)) continue;
+            for (vSphereCloudSlaveTemplate template : ((vSphereCloud) cloud).getTemplates()) {
```
pjdarton, line 44: Rather than casting `((vSphereCloud)cloud)` every time, cast once into a variable and use that.
e.g.
```
if (!(cloud instanceof vSphereCloud)) continue;
vSphereCloud vsCloud = (vSphereCloud) cloud;
for (vSphereCloudSlaveTemplate template : vsCloud .getTemplates()) {
```
Casts are ugly - they're best done as little as possible 😉

pjdarton (changes requested):
We need a unit-test for the new method that's been added to CloudProvisioningAlgorithm (that code can be tested in isolation and has good coverage)

I would _like_ to have a unit-test for the new provisoning code too - some test scenarios where the auto-provisioning is called (I think we can assume Jenkins will call it and don't need to test that Jenkins does so) with different situatuons and check that it decides (correctly) what to provision (stubbing/mocking out the bits that'd talk to a vSphere instance).
e.g. "if we've got no instances, no instance limit and instanceMin is 2 then we expect two instances to be provisioned" and "if we've got 2 instances, no instance limit, and instanceMin is 2 then we expect none to be provisioned" etc.
IME unit-tests are a good way of flushing out "out by 1" mistakes (that are hard to spot "in the real world") and they also help protect functionality from getting broken by later changes (changes done by people who don't use this feature and hence won't notice if they break it _unless_ there's a unit test that tells them)

src/test/java/org/jenkinsci/plugins/NodePreProvisionTest.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    @Test 
+    public void shouldPreProvisionNodesWhenNotEnough() {
+        vSphereCloudSlaveTemplate template = createTemplate(10, 2);
+        provisionNode(template);
+        assertThat(stubVSphereTemplateState.countNodes(), equalTo(1));
+
+        // Here it says that there still should be provisioned 2 nodes, despite the fact there is 1 active already
+        stubVSphereCloud.preProvisionNodes(template);
```
korablin, line 85: The logs: 
```
Jun 01, 2021 5:54:23 PM org.jenkinsci.plugins.vSphereCloud InternalLog
INFO: STARTING VSPHERE CLOUD

Jun 01, 2021 5:54:23 PM org.jenkinsci.plugins.vsphere.tools.CloudProvisioningState logStateChange
WARNING: Marking templateCapacity10instanceMin21 as active : wasPreviouslyPlanned!=true

Jun 01, 2021 5:54:24 PM org.jenkinsci.plugins.vSphereCloud preProvisionNodes
INFO: preProvisionNodesForTemplate(null): should pre-provision 2 nodes

Jun 01, 2021 5:54:24 PM org.jenkinsci.plugins.vSphereCloud calculateMaxAdditionalSlavesPermitted
INFO: There are 0 VMs in this cloud. The instance cap for the cloud is 100, so we have room for more

Jun 01, 2021 5:54:24 PM org.jenkinsci.plugins.vSphereCloud calculateMaxAdditionalSlavesPermitted
INFO: There are 1 VMs in this cloud. The instance cap for the cloud is 100, so we have room for more
```

korablin:
Hey @pjdarton ! Thanks for the comments :) 
I've added a draft code with unit tests for the new provisioning method: "NodePreProvisionTest.java"
But I can't get it working properly, though. Could you help me to understand what is wrong there?

I couldn't find a way to convince `preProvison` method that there is something provisoned already.

src/test/java/org/jenkinsci/plugins/NodePreProvisionTest.java
korablin: Note,  that when it preProvision is invoked it actually sees that "There are 1 VMs in this cloud", which was "provisioned" by the method just a moment ago.
So, I would love to achieve the same state without calling preProvision(); I've tried marking the associated record as `provisionedSlaveNowActive` or `provisioningStarted`, but that doesn't do the trick, neither calling `VSpherePlannedNode.createInstance(stubVSphereTemplateState, nodeName, provisionable);`

korablin:
Hey @pjdarton! If I remove "NodePreProvisionTest.java" do you think this could be landed?

PR #129, opened 2021-09-11
Stop using deprecated `Util#join`

basil:
This method was deprecated in jenkinsci/jenkins#5467, so migrate away from it in favor of native Java Platform functionality.


pjdarton:
I'm going to hold off merging this until https://github.com/jenkinsci/docker-plugin/pull/855 has proved it is safe.

The docker-plugin has much better unit-test coverage (its unit-tests can test that it can spin up new containers, which is possible as the Jenkins CI environment has docker) than this plugin (the Jenkins CI environment doesn't have a vSphere server to test against, so this plugin can't test things "for real") and it's currently failing on the docker-plugin's equivalent PR - I suspect that if the unit-tests were adequate in this plugin, it'd fail here too.

PR #131, opened 2022-05-09
[JENKINS-68477] Prepare vSphere for removal of JAXB and Java 11 requirement

basil:
See [JENKINS-68477](https://issues.jenkins.io/browse/JENKINS-68477). This plugin bundles `com/vmware/vim25/ws/XmlGen.class`, `org/doublecloud/ws/util/ReflectUtil.class`, and Dom4j, all of which consume JAXB. But when running on Java 9+, JAXB is not included on the classpath by default. The only way for a plugin to have access to JAXB when running on Jenkins 2.164 or later on Java 11 is to declare a plugin-to-plugin dependency on JAXB API plugin (recommended) or to embed JAXB into its `.jpi` file. This PR does the former to ensure that vSphere Cloud always has access to JAXB on its classpath. Since JAXB API plugin requires 2.263 or newer, we bump the minimum Jenkins version accordingly. That, in turn, requires updating the plugin parent POM to 4.31 or later and the plugin BOM to a recent version. CC @pjdarton

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>4.14</version>
+        <version>4.31</version>
     </parent>
 
     <artifactId>vsphere-cloud</artifactId>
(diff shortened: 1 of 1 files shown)


Peter-Darton-i2:
@basil Merged - does this warrant doing an immediate release or is it likely that there will be other inbound changes that I should wait for?

basil:
Thanks! I have not tested this plugin with Java 11 (and do not have a vSphere environment to test it with), so I cannot confirm that this PR is sufficient to support Java 11, I do know that this fix is necessary to support Java 11. If you have the time and interest in testing this plugin with Java 11, that would be ideal. However, in the meantime, releasing this PR could not possibly hurt, and may be sufficient for Java 11 support. Jenkins core is planning on requiring Java 11 in weeklies in about a month, and in the September LTS release.

PR #181, opened 2026-07-27
[JENKINS-76498] Fix NullPointerException when reverting to a VM snapshot

ijwalker:
# JENKINS-76498 — Fix NullPointerException reverting to a VM snapshot (+ optional "suppress power on")

https://issues.jenkins.io/browse/JENKINS-76498

## Problem
Reverting a VM to a **named** snapshot throws a `NullPointerException` before the
revert is performed. This affects both the "Revert to Snapshot" build step and the
cloud-agent path that reverts a template snapshot when a vSphere agent launches, so
snapshot-based workflows fail outright.

- **What:** revert a VM to a named snapshot (build step, or agent revert-on-launch).
- **How it fails:** the build fails during the snapshot revert. Pipeline console output:
  ```
  [vSphere] Reverting to snapshot "clean" for VM <vm-name>...
  [Pipeline] End of Pipeline
  ERROR: Cannot invoke "java.lang.Boolean.booleanValue()" because "suppressPowerOn" is null
  Finished: FAILURE
  ```
  The JDK "Helpful NullPointerException" message (JEP 358) names the culprit exactly:
  the boxed `Boolean suppressPowerOn` is `null` and is being unboxed via `booleanValue()`.
- **Expected:** the VM reverts to the named snapshot and the build/agent continues.

## Root cause
Exposed by the yavijava 6.0.05→9.0.2 upgrade (commit d1fbc74): the generated VimPortType.revertToSnapshot_Task binding changed its suppressPowerOn parameter from a nullable Boolean to a primitive boolean, so the SDK now unboxes it unconditionally and the plugin's long-standing single-arg revertToSnapshot_Task(null) calls throw an NPE.

The plugin calls `VirtualMachineSnapshot.revertToSnapshot_Task(HostSystem)` — the
single-argument overload in the bundled yavijava SDK. That overload delegates to
`revertToSnapshot_Task(HostSystem, Boolean suppressPowerOn)` passing a **null**
`Boolean`. During SOAP serialization the boxed `Boolean` is unboxed to a primitive,
and unboxing `null` throws the NPE. By contrast `revertToCurrentSnapshot_Task` takes a
primitive `boolean`, which is why "revert to current snapshot" was unaffected while
"revert to named snapshot" failed.

## Fix
Pass an explicit non-null `Boolean` (`false`, the historical default) at every call site:
- `tools/VSphere.java` — `revertToSnapshot(...)` now calls
  `revertToSnapshot_Task(host, Boolean.valueOf(suppressPowerOn))`.
- `vSphereCloudLauncher.java` — the two agent revert-on-launch call sites now pass
  `Boolean.FALSE`.

`false` means "restore the power state recorded in the snapshot" — exactly the prior
behavior, so existing users see no change.

## Enhancement (opt-in, default off)
Since the SDK already supports it, this also exposes `suppressPowerOn` as an optional
parameter of the "Revert to Snapshot" build step:
- `builders/RevertToSnapshot.java` — new `suppressPowerOn` field via `@DataBoundSetter`
  (the `@DataBoundConstructor` signature is unchanged).
- `RevertToSnapshot/config.jelly` — a checkbox for the field.
- `RevertToSnapshot/help-suppressPowerOn.html` — help text describing the field and its
  default.

When enabled, the VM is left powered off after the revert regardless of the power state
captured in the snapshot.

## Backwards compatibility
- No constructor signatures changed; existing job config and pipeline scripts load as-is.
- Default (`suppressPowerOn = false`) is identical to prior releases.
- No breaking changes.

## Tests
Adds `builders/RevertToSnapshotTest.java` — three unit tests that do **not** require a
live vSphere (they exercise the pipeline data-binding path via `DescribableModel`):
1. `suppressPowerOn` is recognized as a data-bound parameter (guards the
   "Unknown parameter" regression).
2. Omitting the field defaults to `false`.
3. Setting the field to `true` is honored.

`mvn -Dtest=RevertToSnapshotTest test` → `Tests run: 3, Failures: 0, Errors: 0`.

## Environment reproduced on
- Jenkins core: 2.568.1 (LTS, JDK 21)
- Plugin version: 4.530.v1b_68fb_22777c

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/vSphereCloudLauncher.java
+++ b/src/main/java/org/jenkinsci/plugins/vSphereCloudLauncher.java
@@ -197,7 +197,7 @@ public class vSphereCloudLauncher extends DelegatingComputerLauncher {
                         }
 
                         vSphereCloud.Log(slaveComputer, taskListener, "Reverting to snapshot:" + snapName);
-                        Task task = snap.revertToSnapshot_Task(null);
+                        Task task = snap.revertToSnapshot_Task(null, Boolean.FALSE);
                         if (!task.waitForTask().equals(Task.SUCCESS)) {
                             throw new IOException("Error while reverting to virtual machine snapshot");
                         }
@@ -504,7 +504,7 @@ public class vSphereCloudLauncher extends DelegatingComputerLauncher {
             }
 
             vSphereCloud.Log(slaveComputer, taskListener, "Reverting to snapshot:" + snapName);
-            Task task = snap.revertToSnapshot_Task(null);
+            Task task = snap.revertToSnapshot_Task(null, Boolean.FALSE);
             if (!task.waitForTask().equals(Task.SUCCESS)) {
                 throw new IOException("Error while reverting to virtual machine snapshot");
             }
--- a/src/main/java/org/jenkinsci/plugins/vsphere/builders/RevertToSnapshot.java
+++ b/src/main/java/org/jenkinsci/plugins/vsphere/builders/RevertToSnapshot.java
@@ -34,6 +34,7 @@ import org.jenkinsci.plugins.vsphere.tools.VSphereException;
 import org.jenkinsci.plugins.vsphere.tools.VSphereLogger;
 import org.kohsuke.stapler.AncestorInPath;
 import org.kohsuke.stapler.DataBoundConstructor;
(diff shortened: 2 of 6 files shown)


jimklimov (approved):
Looks great, thanks for the catch and fix!

PR #190, opened 2026-08-24
Fix the remaining null-unboxing NPEs

mschneid:
# Fix the remaining null-unboxing NPEs

Follow-up to #189 (JENKINS-76493) and #181 (JENKINS-76498). Both of those were one shape of bug — a boxed
value that is `null` reaching an implicit unboxing — so rather than guess where else it hides, I enumerated
them: every class in `target/classes` disassembled with `javap -c -l`, each
`booleanValue/intValue/longValue/doubleValue` call mapped back to its source line through the
LineNumberTable. That gives 36 unboxing sites, 33 of which are properly guarded. These are the three that
are not.

None of them come from the yavijava 6.0.05 → 9.0.2 upgrade: of the two SDK overloads that delegate `null`
into a primitive stub parameter (`VirtualMachineSnapshot.removeSnapshot_Task(boolean)` and
`revertToSnapshot_Task(HostSystem)`), the plugin calls neither any more after #181 and #189. These three are
the plugin's own optional values.

## 1. `ExposeGuestInfo.waitForIp4` — reachable from a pipeline today

`waitForIp4` is a `Boolean` that `exposeInfo` unboxes unguarded, so the step throws before doing any work
whenever the flag was never set:

```
Cannot invoke "java.lang.Boolean.booleanValue()" because "this.waitForIp4" is null
```

That happens for a pipeline step declaring only the fields it cares about —
`vSphere buildStep: [$class: 'ExposeGuestInfo', vm: '…', envVariablePrefix: '…'], serverName: '…'`, where
`structs` passes `null` for an omitted `Boolean` — for a job configured before the field was introduced
(5a1eca0), and for a JCasC definition that leaves it out.

Read through `Boolean.TRUE.equals` so an unset flag means "do not wait": that is how the step behaved
before the field existed, and it keeps a caller that never asked to wait out of the (unbounded) IP wait
loop.

The step also had **no getter at all** for the field, so `config.jelly`'s `field="waitForIp4"` had nothing
to bind against — the form could not show the stored value and JCasC could not export it. This adds
`isWaitForIp4()`, and turns the checkbox's hard-coded `checked="true"` into `default="true"`, which keeps
the same default for new steps while letting an existing configuration show what it actually has. Without
that, every save silently switched the flag back on.

Covered by a new `ExposeGuestInfoTest` in the style of `RevertToSnapshotTest`: the parameter stays
data-bound, an omitted flag reads as `false`, an explicit `true` is honoured.

## 2. `ReconfigureDisk.selectUnitNumber` — unboxing an optional device key

`vmDevice.getControllerKey() == controller.getKey()` compares a boxed `Integer` that the API documents as
optional against the controller's primitive `int`, so the comparison unboxes whatever the device reports.
The guard in front of it only checks `getUnitNumber() != null`, so a device that has a unit number without
belonging to a controller fails the whole disk reconfiguration. Comparing from the controller's own key
(`Integer.valueOf(controller.getKey()).equals(...)`) reads a missing device key as "not this controller".

Not reproduced against a live VM — reported as a wrong guard, not as an observed failure.

## 3. `vSphereCloudLauncher` — unboxing `waitForVMTools` on launch

Same shape: a boxed `Boolean` that `readResolve()` never defaults, unboxed while launching an agent, so an
agent configuration saved before the field existed cannot start. `overrideLaunchSupported`, two accessors
down, already guards for exactly this case in `isLaunchSupported()`; this does the same. The public
`getWaitForVMTools()` accessor is left as-is so JCasC round-tripping does not change.

## Testing

- `mvn package` on Temurin 21: `BUILD SUCCESS`, 165 tests green (162 existing + 3 new).
- Re-ran the `javap` scan afterwards: the three sites are gone, and the unboxings that remain in those
  classes are the guarded one in `isLaunchSupported()` and `ReconfigureDisk`'s private `Integer retry`
  parameter, which every caller passes a literal to.
- The `ReconfigureDisk` and launcher paths need a live vCenter and an old agent config respectively, so
  they are argued from the types rather than covered by a test.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/vSphereCloudLauncher.java
+++ b/src/main/java/org/jenkinsci/plugins/vSphereCloudLauncher.java
@@ -216,7 +216,8 @@ public class vSphereCloudLauncher extends DelegatingComputerLauncher {
                             break;
                     }
 
-                    if (waitForVMTools) {
+                    // Null in agent configurations saved before the field existed, as with overrideLaunchSupported
+                    if (Boolean.TRUE.equals(waitForVMTools)) {
                         vSphereCloud.Log(slaveComputer, taskListener, "Waiting for VMTools");
 
                         Calendar target = Calendar.getInstance();
--- a/src/main/java/org/jenkinsci/plugins/vsphere/builders/ReconfigureDisk.java
+++ b/src/main/java/org/jenkinsci/plugins/vsphere/builders/ReconfigureDisk.java
@@ -238,8 +238,10 @@ public class ReconfigureDisk extends ReconfigureStep {
 		map.put(7, true); // Unit number 7 is reserved for the controller
 
 		for (VirtualDevice vmDevice : vm.getConfig().getHardware().getDevice()) {
+			// getControllerKey() is optional in the API, so compare from the controller's own key to keep a
+			// device without one from unboxing null
 			if (vmDevice.getUnitNumber() != null &&
-					(vmDevice.getControllerKey() == controller.getKey() || vmDevice.getKey() == controller.getKey())) {
+					(Integer.valueOf(controller.getKey()).equals(vmDevice.getControllerKey()) || vmDevice.getKey() == controller.getKey())) {
 				map.put(vmDevice.getUnitNumber(), true);
 			}
 		}
--- a/src/main/java/org/jenkinsci/plugins/vsphere/builders/ExposeGuestInfo.java
+++ b/src/main/java/org/jenkinsci/plugins/vsphere/builders/ExposeGuestInfo.java
@@ -60,6 +60,12 @@ public class ExposeGuestInfo extends VSphereBuildStep implements SimpleBuildStep
         return envVariablePrefix;
     }
 
+    public boolean isWaitForIp4() {
+        // Null whenever the flag was never set: a pipeline step that omits it, or a job configured before
+        // the field existed. Waiting was not an option back then, so that is the compatible answer.
+        return Boolean.TRUE.equals(waitForIp4);
+    }
+
     @Override
     public String getIP() {
         return IP;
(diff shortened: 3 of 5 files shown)


src/main/java/org/jenkinsci/plugins/vsphere/builders/ReconfigureDisk.java
```diff
@@ -238,8 +238,10 @@ private int selectUnitNumber(VirtualMachine vm, VirtualController controller) {
 		map.put(7, true); // Unit number 7 is reserved for the controller
 
 		for (VirtualDevice vmDevice : vm.getConfig().getHardware().getDevice()) {
+			// getControllerKey() is optional in the API, so compare from the controller's own key to keep a
+			// device without one from unboxing null
 			if (vmDevice.getUnitNumber() != null &&
-					(vmDevice.getControllerKey() == controller.getKey() || vmDevice.getKey() == controller.getKey())) {
+					(Integer.valueOf(controller.getKey()).equals(vmDevice.getControllerKey()) || vmDevice.getKey() == controller.getKey())) {
```
jimklimov, line 244: Just in case: Shouldn't the second `controller.getKey()` comparison be wrapped here too?

src/main/java/org/jenkinsci/plugins/vSphereCloudLauncher.java
```diff
@@ -216,7 +216,8 @@ public void launch(SlaveComputer slaveComputer, TaskListener taskListener)
                             break;
                     }
 
-                    if (waitForVMTools) {
+                    // Null in agent configurations saved before the field existed, as with overrideLaunchSupported
+                    if (Boolean.TRUE.equals(waitForVMTools)) {
```
jimklimov, line 220: Why not introduce a method like `isWaitForIp4()` above, to be consistent with this fix?

jimklimov (approved):
Generally looks good, posted a couple of ideas/questions about maybe making it even better.

jimklimov:
Thanks for the good catches!

src/main/java/org/jenkinsci/plugins/vsphere/builders/ReconfigureDisk.java
jimklimov: My bad, misread the comment above.

